PR #2, opened 2017-07-26
fix(asammdf): add compatibility with python2.x

JulienGrv:
Hi Daniel,

I thought it would be nice to make your module compatible with Python 2.7.
So I did a few changes and it seems to work now but could you make some test and review my changes?
I suggest you create a develop branch on your repository, to keep the master branch clean. The master branch would be only used for release.

Changes:
- Removed the * you have in your get and remove functions. Why are you using this?
- Replaced every super() with super(\<ClassName\>, self) for backward compatibility with Python2.7
- Replaced every iter_unpack with unpack in a loop (not sure if it works!!)

Cheers, Julien

### DIFF
--- a/asammdf/mdf.py
+++ b/asammdf/mdf.py
@@ -54,13 +54,13 @@ class MDF(object):
 
     def __setattr__(self, attr, value):
         if attr == 'file':
-            super().__setattr__(attr, value)
+            super(MDF, self).__setattr__(attr, value)
         else:
             setattr(self.file, attr, value)
 
     def __getattr__(self, attr):
         if attr == 'file':
-            return super().__getattr__(attr)
+            return super(MDF, self).__getattr__(attr)
         else:
             return getattr(self.file, attr)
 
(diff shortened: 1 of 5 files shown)


danielhrisca:
I not against python2 compatibility if this does not interfere with the performance and clarity of the code.

The "*" in the function arguments will gather all positional arguments, and in this way the next arguments will only be passed as named arguments; basically you have to know exactly that you want to use those arguments.

Going from _iter_unpack_ to _unpack_ will require some changes.

I will look into this as soon as I'll get some free time.

danielhrisca:
The module heavily relies on the bytes built-in from python 3 (__bytes__ method of all blocks). Python 2 lacks this type. This is a major blocker for supporting python 2 as I see it now.

danielhrisca:
There were a lot of small changes to be made for Python 2.7 support but I was finally able to run all benchmarks with both Python 3.6 and 2.7 without error. 

I started with a copy of the code from your python 2,x branch and build from there. 

Thank you for the push

danielhrisca:
Solved

Issue #21, opened 2017-12-12
Cannot get signal in MDF4 because of error in dependency_list


JulienGrv:
Hi everyone,

When I try to get a list of signals that belongs to the same ChannelArrayBlock, the last signal has its channel_dependencies not empty (although it is a normal Channel).
The dependency_list being not empty, it tries to get every signals like if it was a ChannelArrayBlock. This signal being itself in the dependency_list, we enter into an infinite loop and asammdf crashes when reaching Python maximal recursion limit.

When looking a bit more in detail, the dependency_list that we get for this signal, is actually the dependency_list of the ChannelArrayBlock it belongs to.

In the ChannelArrayBlock I tested, I have 3 signals, so when I put an offset of three, the get function is working again:
```
dependency_list = grp['channel_dependencies'][ch_nr + 3]
instead of:
dependency_list = grp['channel_dependencies'][ch_nr]
in get function
```

However, I am not sure yet if this offset is constant or if it depends.

I am still doing some investigation to find out why the channel_dependencies does not match with the channels we have.

danielhrisca:
Hello Julien,

I'm not sure what this kind of recursion might mean; could you post a MDFValidator screenshot of this Channel and its ChannelDepenedencyBlock?

If you're allowed to send the file via e-mail it would be perfect for investigation.

danielhrisca:
I will probably disappoint you: I often use _print_, _pprint_ and _hex_; most times this is enough if you know where to place them. We also have to thank _Vector Gmbh_ for _MDFValidator_. 
I should definitely learn to use the debuggers more.

JulienGrv:
Ah ah, I am not disappointed at all. Good programmers always know directly where the error comes from, no need of a debugger ;) 

Anyway, I will use the MDFValidator to see if it gives me some more information to find my way in this big MDF file!

JulienGrv:
In theory, I should have as many channels as the sum of the channels_dependencies for every groups?

In my case, there is 291081 channels in the channels_db. While the sum of channels_dependencies for every groups is 335095:

```python
lengths = [len(grp['channel_dependencies']) for grp in self.groups]
total = sum(lengths)
print(total)  # prints 335095
print(len(self.channels_db))  # prints 291081
```

danielhrisca:
If you have repeating channel names then you could see a difference.

Try this:

``` python
print(len(self.channels_db))

lengths = [len(grp['channel_dependencies']) for grp in self.groups]
[...]

PR #23, opened 2017-12-14
fix(mdf4): error maximum recursion exceeded

JulienGrv:
Fix issue #21

### DIFF
--- a/asammdf/mdf4.py
+++ b/asammdf/mdf4.py
@@ -640,6 +640,8 @@ class MDF4(object):
                 stream.seek(channel['component_addr'], v4c.SEEK_START)
                 blk_id = stream.read(4)
                 if blk_id == b'##CN':
+                    index = len(grp['channel_dependencies'])
+                    grp['channel_dependencies'].append(None)
                     ch_cntr, composition = self._read_channels(
                         channel['component_addr'],
                         grp,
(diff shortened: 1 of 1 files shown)


JulienGrv:
That's right, I will edit this

JulienGrv:
I can do a copy of ch_cntr like this:

```python
from copy import copy

[...]

index = copy(ch_cntr)
grp['channel_dependencies'].append(None)
ch_cntr, composition = self._read_channels(
    channel['component_addr'],
    grp,
    stream,
    dg_cntr,
    ch_cntr,
    True,
)
grp['channel_dependencies'][index] = composition
```

I could also user insert() instead of appending None, what do you think is most suitable?

danielhrisca:
just a simple ``index = ch_cntr `` is enough.

JulienGrv:
No, I tried and it does not work unfortunately.

danielhrisca:
It should be ``index = ch_ntr - 1`` because _ch_cntr_ is incremented just before the _if_ statement.

JulienGrv:
Oh! Alright! and insert?

danielhrisca:
I think it's better to leave it like it is with appending.

danielhrisca:
Thank you Julien for fixing this subtle bug.

Issue #14, opened 2017-10-25
Repo clean-up


danielhrisca:
It is advised to avoid adding large binary files to a git repository.

I've broken this rule and the result is clear:
- a clone of _asammdf_ takes ~118MB for ~1900 objects
- as a comparison the _CPython_ has ~216MB for ~680000 objects

![image](https://user-images.githubusercontent.com/20952040/31994708-50ec536a-b98a-11e7-99c7-e18cdbc375e6.png)
![image](https://user-images.githubusercontent.com/20952040/31994712-558a131c-b98a-11e7-8b83-5af3b98123a8.png)

This is why I've decided to clean-up the repository and remove the bloat.

The test files used for bench-marking will be removed, as well as the PNG results. I can provide the test files upon request to whoever is interested.  

I know this will break the existing forks and I apologize for it.

danielhrisca:
Here are the files used for bench-marking:

[test files.zip](https://github.com/danielhrisca/asammdf/files/1416933/test.files.zip)

PR #27, opened 2017-12-14
pull CI improvements from Mihai

danielhrisca:


yahym:
It seems that with python 3.4 there is a failure with linux package on travis. I did not try it on windows py3.4 yet.

```
======================================================================
ERROR [0.007s]: test_convert_from_mdf3_20_to_mdf4_10 (test.test_mdf.TestMDF)
----------------------------------------------------------------------
  File "/home/travis/build/danielhrisca/asammdf/test/test_mdf.py", line 35, in test_convert_from_mdf3_20_to_mdf4_10
    converted = meas.convert(to='4.10', memory='minimum')
  File "/home/travis/build/danielhrisca/asammdf/asammdf/mdf.py", line 173, in convert
    sigs.append(self.get(group=i, index=j, data=data))
  File "/home/travis/build/danielhrisca/asammdf/asammdf/mdf3.py", line 2388, in get
    record = fromstring(data, dtype=dtypes)
  File "/home/travis/virtualenv/python3.4.6/lib/python3.4/site-packages/numpy/core/records.py", line 709, in fromstring
    _array = recarray(shape, descr, buf=datastring, offset=offset)
  File "/home/travis/virtualenv/python3.4.6/lib/python3.4/site-packages/numpy/core/records.py", line 423, in __new__
    strides=strides, order=order)
TypeError: 'float' object cannot be interpreted as an integer
```

JulienGrv:
Hi! 

Thank you for adding unit tests!
I saw you uploaded some MDF files directly in the repository. I think it would be better to store them somewhere else and download them as needed. Otherwise we will have the same problem as in issue #14

danielhrisca:
We'll have to go with autogenerated files or store them somewhere else (inside an issue maybe)

yahym:
- I really appreciate you took a look.... my idea is to have a continuous test prove, 
before deploying this package inside an application.

- Unit test is good only if you focus on validating the package/features and not the code itself.
- Still, from my point of view we should also consider a standard[complex] `mdf` generated/saved 
from an official/widely used tool - that is why I generated for each mdf version one file with one channel 
to keep a small size, and I am not really happy because is reading just a basic file for now.
(we can keep it also elsewhere and not in the repo).

- Also as you saw, I tried to use the examples provided, and to avoid reinventing the wheel.

- If I sparked your interest and we have the same goal in the end, I will consider your proposal and find another way of handling the test vectors.

I am really confident in the potential of this package and I hope we can make it better.


As about py34 failure, it beats me... So if you have any idea please fill free to guide me :)

JulienGrv:
For MDF4, we can use the [example file](https://www.asam.net/index.php?eID=dumpFile&t=f&f=199&token=fffa2c6f1b7d33bc599abe7f9d232c5846813669) given by ASAM at the end of this page: https://www.asam.net/standards/detail/mdf/wiki/

danielhrisca:
I've added a zip file to issue #29 that contains the example file converted to different version using CANape. I'm working on using this archive for unittest. This should be ready in a few days.

yahym:
Not very clear for me.... 
Check commit with `- `set -e` is used to fail the build immediate` - `set -e` will fail the build if any of the scripts exits with 1, if you do not use this than the build will be passed :(. 

It seems that .travis should fail according to [customizing-the-build](https://docs.travis-ci.com/user/customizing-the-build/)


> Breaking the Build #
> If any of the commands in the first four stages of the build lifecycle return a non-zero exit code, the build is broken:
> 
> If before_install, install or before_script return a non-zero exit code, the build is errored and stops immediately.
> If script returns a non-zero exit code, the build is failed, but continues to run before being marked as failed.
> The exit code of after_success, after_failure, after_script and subsequent stages do not affect the build result. However, if one of these stages times out, the build is marked as a failure.
>

yahym:
ZipFile in python 27 does not know to extract 7zip archives ([lzma](https://en.wikipedia.org/wiki/Lempel%E2%80%93Ziv%E2%80%93Markov_chain_algorithm) compreassion)

danielhrisca:
In this case we needed to compress the files with a compression type compatible with zip file. No need to bring in pyunpack for a few kB of saving in file size

yahym:
I run the tests locally in py27 with a regular compressed file and tests are passed with your initial implementation... so, now, waiting for a new `test.files.zip`.

yahym:
All is needed for codacy coverage is to include to travis-ci account, under project settings, an env variable `CODACY_PROJECT_TOKEN` (which can be taken from codacy account, when configuring `coverage`)

e.g. [coverage on yahym account](https://www.codacy.com/app/yahym/asammdf/file/13096703992/coverage?bid=6045316&fileBranchId=6045316)

danielhrisca:
If I understand this right, then it might not be possible to use he cadacy coverage from a PR:
https://github.com/travis-ci/travis-ci/issues/7412

yahym:
You need to follow these steps

1. On your [codacy dashboard](https://www.codacy.com/app/danielhrisca/asammdf/dashboard)
2. Select `Setup Coverage`
3. Copy `codacy token` xxxxx
4. Go to [travis-ci](https://travis-ci.org/danielhrisca/asammdf)
5. Go to `More Options` -> `Settings`
6. Add new environment variable `CODACY_PROJECT_TOKEN` and paste the `codacy token` 
    - make sure `Display value in build log` is set to `Off`

7. Update markdown badge to `README.md`
    - Take the coverage badge from your [codacy Settings](https://www.codacy.com/app/yahym/github_publish/settings)
    - Make a push with new `README.md` to trigger a build
8. Done. You should have the coverage added as a badge from your project (which makes more sense than from mine :)).


> Interesting though: Is not clear for me: If you have multiple python environments 
> which coverage will be published (I think the last one that runs - not fun - you will never have 100% coverage like this with py27 and py34 implementation)
> 

If you follow the steps above, also the proposed PR #27 should pass (all is needed is env var `CODACY_PROJECT_TOKEN` added to `travis-ci` to your `asammdf` project and a push)

danielhrisca:
I have set up the environment variable for codacy, and if you look at Travis build #53 it is passed and you see the variable in the log. The problem is that when the build is triggered by an PR from a fork then the variable is not used in the build.

yahym:
finally... thank you for your patience and support...

PR #28, opened 2017-12-14
Bugfix/append

JulienGrv:


JulienGrv:
But the problem here is that if you pass a list with several channels + a channel which is actually a composed channel of some channels already in this list, we end up with undesired duplicates.

Step by step:
First all the channel without dependencies will be appended, then the channel with dependencies will be appended, and finally all its channel dependencies, which are in this case already there and will be appended twice. This behavior is not expected.

I think we should also look for a better concept of handling channel with dependencies. I think we should try to recreate the tree of the original MDF if it is possible.

danielhrisca:
Duplication is partially handled by the __excluded_channels_ method of _MDF_ class, but only in case of converting, merging and re-sampling. 

One way to handle this is to compare the samples of simple channels and composition fields that share the same name, but this must be done first thing in the append method since the composed channels form a linked list, so a simple channels should be excluded in this case and only the composed field must be appended.

JulienGrv:
OK I see. So I think I have to handle this problem before append in the filter method, maybe with the help of this _excluded_channels method. We will see.

danielhrisca:
Hello Julien,

have a try with this development commit: https://github.com/danielhrisca/asammdf/commit/abf6ed331a08a4b04444a3ae48790ae42cb3aa56

JulienGrv:
It's done. Could you make a release after merging this fix? Thank you!

danielhrisca:
I have some work on cleaning up the initial unittest suite. I think it will be ready in a couple of days.

PR #37, opened 2018-01-30
Fix signed non-byte-sized integers in MDF3

jackjweinstein:
The MDF3.get method was masking irrelevant bits from non-byte-sized
integers and then relying on numpy.ndarray.astype to convert back to
signed. Since numpy only has byte-sized data types, this meant that the
wrong two's complement was used (e.g. 2**16 rather than 2**14 for a
14-bit integer encoded in a np.uint16).

This change computes two's complement for non-byte-sized signed integers
using numpy.where.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -158,72 +158,76 @@ Files used for benchmark:
 ================================================== ========= ========
 Open file                                          Time [ms] RAM [MB]
 ================================================== ========= ========
-asammdf 2.8.1 full mdfv3                                1259      260
-asammdf 2.8.1 low mdfv3                                 1076      106
-asammdf 2.8.1 minimum mdfv3                              767       52
-mdfreader 2.7.3 mdfv3                                   3146      392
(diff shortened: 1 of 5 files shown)


danielhrisca:
Hello Jack,

thank you for your work!

Before merging the changes there are some details to take care of:
1. the _master_ branch should hold the release code (whatever is the latest on PyPI). Please make the PR on the _pre_generator_ branch; this is the latest code base similar to the 2.8.1 release before the many changes coming for the 3.0.0 release.
2. The same applies for the_MDF4_ get method, and it should also be updated. Since you have already created a staticmethod it will be easy to move it to the _utils.py_ file to share it on all MDF versions.
3. throughout the code we have used numpy style docstrings. Please use this style as well to keep it consistent.

jackjweinstein:
Okay! I've just moved my work to pre_generator; I'm going to close this PR.

PR #38, opened 2018-01-30
Use numpy.where for signed non-byte-sized integers

jackjweinstein:
The MDF3.get method was in master was masking irrelevant bits from
non-byte-sized integers and then relying on numpy.ndarray.astype to convert
back to signed. Since numpy only has byte-sized data types, this
meant that the wrong two's complement was used (e.g. 2**16 rather than
2**14 for a 14-bit integer encoded in a np.uint16).

This change computes two's complement for non-byte-sized signed integers
using numpy.where.


danielhrisca:
Thank you and we'll one to the contributors team!

PR #46, opened 2018-03-02
Include LICENSE in MANIFEST.in

isuruf:
So that LICENSE is packaged in the source distribution


danielhrisca:
Thank you for the support!

PR #49, opened 2018-03-04
Update readme for conda-forge package installation

fkohlgrueber:


danielhrisca:
Is it the preferred way to use con-forge instead of anaconda-cloud?

fkohlgrueber:
I think so. The advantage of conda-forge is that it's community-based so that fixing project builds when the maintainer isn't responsive anymore is possible. Another advantage is that companies might have popular channels such as conda-forge mirrored in their network so that people with restricted internet access can still install packages from them. Other channels on anaconda cloud might not be accessible from within the companies network.

PR #57, opened 2018-04-03
fix(mdf_v4): catch parser error if comment is not an xml

JulienGrv:


asammdf/mdf_v4.py
```diff
@@ -433,24 +433,28 @@ def _read(self):
...
+                                grp['can_id'] = can_id
+                                grp['message_name'] = message_name
+                                grp['message_id'] = message_id
 
-                        else:
-                            warnings.warn('Invalid bus logging channel group metadata: {}'.format(comment))
+                            else:
+                                warnings.warn('Invalid bus logging channel group metadata: {}'.format(comment))
+                                channel_group['flags'] &= ~v4c.FLAG_CG_BUS_EVENT
+                        except ET.ParseError as e:
```
danielhrisca: I would leave out the "as e" assignment if it's not used in the exception block.

asammdf/mdf_v4.py
```diff
@@ -433,24 +433,28 @@ def _read(self):
...
-                                name = e.get('name')
-                                if name == 'MessageID':
-                                    if e.get('ci') is not None:
-                                        can_id = int(e.get('ci'))
-                                    message_id = int(e.text)
-                            grp['can_id'] = can_id
-                            grp['message_name'] = message_name
-                            grp['message_id'] = message_id
+                        try:
+                            comment_xml = ET.fromstring(comment)
```
danielhrisca: At this point the channel group comment should be a valid XML with relevant information about the CAN message, or the bus logging is done for another bus than CAN.

I guess you have already a measurement, so please print:
```python
print(channel_group.acq_name, channel_group.comment)
print(channel_group.acq_source)
```

JulienGrv:
Bug fixed in development branch

Issue #53, opened 2018-03-08
Is it possible to append a bool signal ?


dglent:
# Python version
('python=3.6.3 (v3.6.3:2c5fed8, Oct  3 2017, 17:26:49) [MSC v.1900 32 bit '
 '(Intel)]')
'os=Windows-7-6.1.7601-SP1'
'numpy=1.13.3'
'asammdf=3.2.1'
# Code 
the existing signal ter15st has logical type in inca: 
>>> mdffile.get('Ter15st')
Signal(name=Ter15st\CCP:1, samples=array([1, 1, 1, ..., 1, 1, 1], dtype=uint8), timestamps=array([   3.1090906 ,    3.1205086 ,    3.1292956 , ...,  245.62956445,
        245.6399234 ,  245.64957463]), unit=-, conversion={'id': b'CC', 'block_len': 62, 'range_flag': 0, 'min_phy_value': 0.0, 'max_phy_value': 0.0, 'unit': b'-\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00', 'conversion_type': 0, 'ref_param_nr': 2, 'b': 0.0, 'a': 1.0}, comment=Terminal 15 status
Terminal 15 status, raw=False)
  ## Code snippet
simt = timestamps array copied form an existing signal
siml = list of values (1, 0) copied from an existing signal
>>> sl = np.asarray(siml, dtype=bool)
>>> sig = asammdf.Signal(sl, simt, name='NewSignal')
>>> sig
Signal(name=NewSignal, samples=array([ True,  True,  True, ...,  True,  True,  True], dtype=bool), timestamps=array([   3.1090906 ,    3.1205086 ,    3.1292956 , ...,  245.62956445,
        245.6399234 ,  245.64957463]), unit=, conversion=None, comment=, raw=False)
>>> len(sl)
24222
>>> len(simt)
24222
>>> mdffile.append([sig,])
  
  ## Traceback
Traceback (most recent call last):
  File "<pyshell#48>", line 1, in <module>
    mdffile.append([sig,])
  File "C:\WinPython-32bit-3.6.3.0Qt5\python-3.6.3\lib\site-packages\asammdf\mdf_v3.py", line 1356, in append
    signal.samples.shape,
[...]

PR #59, opened 2018-04-12
fix(mdf_v4): uint8 data type is converted to bool

JulienGrv:
I saw you convert data to bool when the bit_count is equal to 1. But still, in MDF Validator, the data type is 0 so this should be a uint IMHO.

### DIFF
--- a/asammdf/mdf_v4.py
+++ b/asammdf/mdf_v4.py
@@ -4405,20 +4405,17 @@ class MDF4(object):
                                 ch_nr,
                             )
 
-                    if bits == 1:
-                        vals = array(vals, dtype=bool)
-                    else:
-                        data_type = channel['data_type']
-                        channel_dtype = array(
-                            [],
-                            dtype=get_fmt_v4(
-                                data_type,
-                                bits,
-                                channel_type,
-                            ),
-                        )
-                        if vals.dtype != channel_dtype.dtype:
-                            vals = vals.astype(channel_dtype.dtype)
+                    data_type = channel['data_type']
+                    channel_dtype = array(
+                        [],
+                        dtype=get_fmt_v4(
+                            data_type,
+                            bits,
+                            channel_type,
+                        ),
+                    )
+                    if vals.dtype != channel_dtype.dtype:
+                        vals = vals.astype(channel_dtype.dtype)
 
                     if not samples_only or raster:
                         timestamps.append(self.get_master(gp_nr, fragment))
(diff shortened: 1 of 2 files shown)


danielhrisca:
Hello Julien,

I think we should make this an option for the _configure_ method by taking into account the following:
* there is no dedicated data type for boolean 
* CANape and Inca treat single bit unsigned channels as logical channels (see also issue #53)
* there is added information to single bit channels that is lost if they are treated as uint8 channels (this happens with _cut_, _convert_, _filter_, _concatenate_ and _stack_ methods)
* everybody can choose how single bit unsigned values are treated (bool/uint8)

The best way would probably be to pass the new configuration option to _get_fmt_v3_, _get_fmt_v4_, _fmt_to_datatype_v3_ and _fmt_to_datatype_v4_
 

It is also necessary to implement the same behavior for mdf version 2 and 3 as well.

danielhrisca:
It appears that CANape treats all unsigned channels with bit a bitcount less or equal to 4 as logical channels.

This is how CANape displays a 4bit unsigned channel:

image

and the same channel, with the bit count set to 5:

image

JulienGrv:
Hi Daniel,

CANape treats the channel as a logical channel for the graphic view, but when you look at the property, it is still a uint, isn't it? I think the conversion to bool should be done for the plotting part, but not for the reading part.

> there is added information to single bit channels that is lost if they are treated as uint8 channels (this happens with cut, convert, filter, concatenate and stack methods)
What do you mean by that? What information will be lost?

danielhrisca:
> What do you mean by that? What information will be lost?

I mean that if you see a 1-bit channel you can be sure it is a flag.

What I would like to have is a mechanism to preserve the bit count for unsigned channels with <=8 bits, and the option to return 1-bit channels as numpy bool if the user desires.

JulienGrv:
The option sounds good to me, at least it makes it clear on the user side why that it is an intended behaviour

danielhrisca:
Seeing that you have not made progress in this pull request I went ahead and made an implementation (https://github.com/danielhrisca/asammdf/commit/a4e53524dd457aab5a33d178aee54a560be5a0c0): 
* the default is now to return _uint8_ instead of _bool_
* the user can configure to get _bool_ instead of _uint8_
* the bit count is preserved in case of unsigned channels with bit count <=4

What do you think?

JulienGrv:
It looks perfect! I will close this PR. I can't wait for the next release, this one will probably be my new favourite :)

PR #85, opened 2018-07-31
fix MDF.save comment address missing

venden:
mdf_v3 save_with_metadata lost header block's comment address.


danielhrisca:
Thank you for the fix!

Issue #89, opened 2018-08-21
Creating a MDF instance from a generic file-like object


kasuteru:
This is in direct reference to #71, which fixes the problem for file-like objects of the type BytesIO. 

There are other file-like objects: In my instance, I try to read an object from Microsoft Azure ADLS. The file-like object in this case is of type `azure.datalake.store.core.AzureDLFile` ([API](https://azure-datalake-store.readthedocs.io/en/latest/api.html)). Since #71 uses `isinstance(name, BytesIO)` to check for file-like objects, this fails for data files in the cloud.

Since manually adding all possible file-like object types will be impossible, I propose the solution [that the pandas library uses](https://github.com/pandas-dev/pandas/blob/68273a7a8b05049b311ea207f78e6d6f30c6eaf5/pandas/core/dtypes/inference.py#L158): Replace `isinstance(name, BytesIO)` with _a function similar to the one at the bottom_ of this issue. If you want, I can integrate it myself and send you a PR.

Edit: The function below does not cover it exactly, since asammdf at the moment uses other attributes like `readinto()` as well. I will try to find a solution for this in a fork and update you if I find a solution.

```
def is_file_like(obj):
    """
    Check if the object is a file-like object.

    For objects to be considered file-like, they must
    be an iterator AND have either a `read` and/or `write`
    method as an attribute.

    Note: file-like objects must be iterable, but
    iterable objects need not be file-like.

    .. versionadded:: 0.20.0

    Parameters
    ----------
    obj : The object to check.

    Returns
    -------
    is_file_like : bool
        Whether `obj` has file-like properties.

    Examples
    --------
    >>> buffer(StringIO("data"))
    >>> is_file_like(buffer)
    True
    >>> is_file_like([1, 2, 3])
    False
    """

    if not (hasattr(obj, 'read') or hasattr(obj, 'write')):
        return False

    if not hasattr(obj, "__iter__"):
        return False

    return True
```

kasuteru:
@danielhrisca : I forked the code and modified it to handle arbitrary file-like objects. Unfortunately, [this line](https://github.com/danielhrisca/asammdf/blob/7a7ff6e21f5c7f5f0c582816fdf8f2dd5a5baeea/asammdf/mdf_v4.py#L1244) uses a function that is apparently not always implemented in file-like objects. Is there a workaround for this line that can be used if readinto is not implemented? tbh, I do not exactly understand what this line does...

danielhrisca:
[...]

PR #90, opened 2018-08-22
Add support for generic file-like objects

kasuteru:
As discussed in #89 .

I have tested it with some internal MDF4-files for opening by path name, by file object, or by AzureDLFile object. Travis CI fails for me because I do not have Codacy enabled, but passes the other tests before. I think it should pass for this PR. This should be a non-breaking change anyways since the code didn't really change for anything but file-like objects that do not implement the `readinto()` method.

In case anybody stumbles upon this from Google, here is a code snippet to use this with Azure ADLS objects:

```
import asammdf
from azure.datalake.store import core, lib, multithread
 
token = lib.auth() # Authorize with Azure
                  
adlsAccountName = "Name_of_ADLS_Storage_Account"
adlsFileSystemClient = core.AzureDLFileSystem(token = token, store_name=adlsAccountName)

with adlsFileSystemClient.open('path/to/myData.mf4', 'rb') as f:
    mdf_azure = asammdf.MDF(f)
    
```


danielhrisca:
Thank you Thomas for your work!

PR #193, opened 2019-05-27
add scrambling of attachments in mdf.scramble

JulienGrv:


danielhrisca:
Hello Julien,

I think we should also write 0x00 bytes for the embedded attachment data, via a new argument to the function that default to True.

JulienGrv:
Hi Daniel,

Makes sense, how should we name the option? skip_attachments=False ? or maybe just attachments=True ?

danielhrisca:
I think that _skip_attachments=False_ is clear

JulienGrv:
Ok, do you have a measurement sample for me with embedded attachment?

danielhrisca:
```python
from asammdf import MDF

m = MDF()

m.attach(b'sample attachment', 'sample.bin', 'this is a demo attachment')

m.save('attachment.mf4')
```

danielhrisca:
Thank you Julien!

JulienGrv:
You're welcome. Can you do a release? Thanks

PR #200, opened 2019-06-19
[Bugfix] attribute error when CAN database could not be loaded

JulienGrv:
Do not process CAN database if db is None


danielhrisca:
Thank you Julien!

PR #201, opened 2019-06-20
Partial finalization support for asammdf

MatinF:


asammdf/blocks/mdf_v4.py
```diff
@@ -259,55 +259,59 @@ def __init__(self, name=None, version="4.10", **kwargs):
...
-                "in case a VLSD CG block is used"
+                " Update of offset values for VLSD channel required"
+                " in case a VLSD CG block is used"
             )
 
             logger.warning(message)
 
+        return flags == 0
+
     def _read(self, mapped=False):
```
danielhrisca: this will have a big performance impact. 

The best case approach would be to read the and at the end perform the finalization by modifying the loaded file structure (using the blocks defined in v4_blocks). This would allow to read the file once and still maintain good performance, but it need refactoring on your side and also in the asammdf to read all blocks (some blocks are skipped now) to create the mapping of address -> block to finalization steps.

asammdf/blocks/FinalizationShim.py
```diff
@@ -0,0 +1,693 @@
...
+        self.file_location: int = 0
+        self.links: Tuple[int, ...] = (0,) * header.link_count
+        
+        return
+    
+    def load(self, stream: BinaryIO):
+        """
+        Function responsible for loading anything past the header of the block.
+        Expects the stream position to be just after the header.
+        :param stream:      BinaryIO for accessing the block data.
```
danielhrisca: please use numpy style docstrings

asammdf/blocks/FinalizationShim.py
```diff
@@ -0,0 +1,693 @@
+import os
```
danielhrisca: please change the file name to lower case with underscores

asammdf/blocks/v4_blocks.py
```diff
@@ -14,6 +14,7 @@
 from pathlib import Path
 
 import numpy as np
+from asammdf.blocks.utils import is_file_like
```
danielhrisca: a relative import can be used here

asammdf/blocks/mdf_v4.py
```diff
@@ -5958,7 +5958,7 @@ def _finalize(self):
         Attempt finalization of the file.
         :return:    None
         """
-        from .FinalizationShim import FinalizationShim
+        from .finalization_shim import FinalizationShim
```
danielhrisca: I think this import can go at the top of the file

danielhrisca:
Thank you Martin!

PR #210, opened 2019-07-17
fix parse error issue for header comment

JulienGrv:
Hi Daniel,

I had a bunch of measurement with wrong XML syntax in head block comment so I had to implement this fix.

### DIFF
--- a/asammdf/blocks/v4_blocks.py
+++ b/asammdf/blocks/v4_blocks.py
@@ -4496,7 +4496,7 @@ class HeaderBlock:
             try:
                 comment_xml = ET.fromstring(comment)
             except ET.ParseError as e:
-                logger.error("could not parse header block comment; %s", e)
+                logger.error(f"could not parse header block comment; {e}")
             else:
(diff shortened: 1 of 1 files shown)


asammdf/blocks/v4_blocks.py
```diff
@@ -4493,19 +4493,23 @@ def __init__(self, **kwargs):
...
-                    elif name == "department":
-                        self.department = e.text
-                    elif name == "project":
-                        self.project = e.text
-                    elif name == "subject":
-                        self.subject = e.text
+            try:
+                comment_xml = ET.fromstring(comment)
+            except ET.ParseError as e:
+                logger.error("could not parse header block comment; %s", e)
```
danielhrisca: Please use f-strings or .format method. I hate % interpolation with a passion.

JulienGrv: makes sense, I will go for f-strings since it only needs to support python3

danielhrisca:
Thank you Julien!

PR #251, opened 2019-10-29
[bugfix] cannot read channel group properties

JulienGrv:


danielhrisca:
Hello Julien,

we need the channel index to be integer so that we can set the mime data for the drag & drop operation. Please try with the new development branch code to see if you still get errors

JulienGrv:
It worked thx

PR #271, opened 2019-11-27
ignore parsing error of comment in mdf_v4._process_can_logging

JulienGrv:


danielhrisca:
Thank you Julien!

danielhrisca:
What recording application causes this problem?

JulienGrv:
The comment was not of xml format but just a name

JulienGrv:
@danielhrisca the name was something like: CAN2_XXX.CAN_DataFrame

PR #295, opened 2020-01-20
signals and sigs are zip with different length in mdf.select

JulienGrv:


danielhrisca:
Good catch Julien, thank you!

PR #306, opened 2020-02-12
persisting last opened file path in QSettings, connecting 'Ctrl + O' shortcut to Open File

driftregion:


asammdf/gui/widgets/main.py
```diff
@@ -77,6 +77,7 @@ def __init__(self, files=None, *args, **kwargs):
         icon.addPixmap(QtGui.QPixmap(":/open.png"), QtGui.QIcon.Normal, QtGui.QIcon.Off)
         action = QtWidgets.QAction(icon, "Open", menu)
```
danielhrisca: Hello Nick,

please also mark the shortcut next to the action text.

Thank you for the PR!

driftregion: After I added that line, I was surprised to notice that Qt seems to do this automatically.
image

danielhrisca: I was not aware of this. Nice find. It will probably work with the other shortcuts as well

PR #308, opened 2020-02-12
cease explicit propagation of unhandled key combinations to parents

driftregion:
Pressing modifier keys [ctrl, alt] with the plot window open was causing the
following error to appear as a pop-up:
```
no access to protected functions or signals for objects not created from Python
File "/home/njames/repos/asammdf/asammdf/gui/widgets/plot.py", line 499, in keyPressEvent self.parent().keyPressEvent(event)
```

The Qt documentation [suggests](https://doc.qt.io/qt-5/qevent.html#ignore) calling QEvent::ignore() on unhandled keyPressEvents to pass them on to parent widgets.


danielhrisca:
I thin I've used this method on other widgets as well

PR #313, opened 2020-02-24
fix offsets in mdf4 structures with non-byte members

dxpke:
When generating signals for a structure, the designated size in bytes of
an element is continuously accumulated to calculate the offsets of
all elements. When an element does not have a size that is a multiple
of 8 (e.g., a boolean flag), the byte size was calculated to be zero,
which means that the subsequent channel erroneously received the same
offset. Fix by rounding up the byte size.

### DIFF
--- a/asammdf/blocks/mdf_v4.py
+++ b/asammdf/blocks/mdf_v4.py
@@ -4929,7 +4929,7 @@ class MDF4(object):
             if sig_type in (v4c.SIGNAL_TYPE_SCALAR, v4c.SIGNAL_TYPE_STRING):
 
                 s_type, s_size = fmt_to_datatype_v4(samples.dtype, samples.shape)
-                byte_size = s_size // 8
+                byte_size = s_size // 8 or 1
 
                 fields.append(samples)
                 types.append((field_name, samples.dtype, samples.shape[1:]))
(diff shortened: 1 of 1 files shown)


danielhrisca:
Hello @dxpke 
thank you for your interest.

I am not sure though how you can reach that code path and still have a non-standard size array, since at that point you should have a numpy array and numpy does not support this.

Can you share some code to illustrate your use case?

dxpke:
Sure! You hit this when you have e.g. the following structure:

```python
dt = numpy.dtype([('test1', np.bool_), ('test2', np.uint8)])
x = numpy.zeros(5, dtype=dt)
```

NumPy does not have any problem with this. It allocates one byte to `test1` and one byte to `test2`. But if you try to export this array `x` with asammdf, both `test1` and `test2` will get the same offset because of the integer division of the bit size (i.e., `1 // 8` is 0, so offset does not get advanced). It will also give an error in MDF validator due to this.

Let me know if this helps or whether you would need a complete MWE.

danielhrisca:
I see. 

For performance reasons I prepare to have this https://github.com/danielhrisca/asammdf/blob/master/asammdf/blocks/mdf_v4.py#L3085 instaed of _ceil_

```python
In [1]: from math import ceil

In [2]: s_size = 1

In [3]: %timeit ceil(s_size / 8)
184 ns ± 4.04 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)

In [4]: %timeit s_size // 8  or 1
41.6 ns ± 1.39 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
```

Please also add it to this line https://github.com/danielhrisca/asammdf/blob/master/asammdf/blocks/mdf_v4.py#L5362

dxpke:
Makes sense. I'll update accordingly.

danielhrisca:
Thank you @dxpke ! Sorry for the delay!

PR #370, opened 2020-05-28
Make the current version work (again) with unsorted, non-aligned MDF4 files

kopytjuk:
Since >= 5.18.1 opening unsorted, non-byte aligned files was not possible. 

Especially https://github.com/danielhrisca/asammdf/blob/4aa9526246ed414cf96ef80644014c515f0fdb25/asammdf/blocks/mdf_v4.py#L9302 overwrote a variable `size` which was defined in a scope above which caused trouble.

In addition I (re-)added support for channel groups without `cg_flags` set. Why was it removed?


check_file.py
```diff
@@ -0,0 +1,18 @@
+import asammdf
```
danielhrisca, line 1: I don't think we need this file in the package

danielhrisca:
> In addition I (re-)added support for channel groups without `cg_flags` set. Why was it removed?

I can't remember this change. Maybe it was accidental

kopytjuk:
@danielhrisca Daniel, do you know what causes the problem in the Codacy/PR Quality Review check? I get errors in the "diff" and "deltas" check ...

danielhrisca:
Thank you for the contribution Marat!

PR #380, opened 2020-07-06
Patch to avoid error when your enviroment doesn't have pyQT installed

nos86:
moved to development branch


danielhrisca:
Thank you Salvo!

PR #404, opened 2020-08-22
Added warning message to Numeric view for channels that are added tha…

Bartimaeus-:
…t do not have monotonous increasing time stamps. Currently Numeric view will show the last value it finds for a given timestamp, which makes it incompatible for files that have a timestamp jump with more than one value for a given timestamp value. This warning was already present for plot view and having it in Numeric view as well should alert the user that there is possibly something wrong with their file
[timeJump.zip](https://github.com/danielhrisca/asammdf/files/5112959/timeJump.zip)


danielhrisca (approved):
Looks good, thank you!

PR #410, opened 2020-09-11
add raw argument to mdf_v4.py

tobiasandorfer:
add raw argument to "get_can_signal" funktion, in mdf_v4.py file, default value is False


danielhrisca:
Thank you Tobias!

PR #423, opened 2020-09-26
Improve handling of channel group acq_name, acq_source and comment

JulienGrv:
Requires PR #421

### DIFF
--- a/asammdf/blocks/mdf_v3.py
+++ b/asammdf/blocks/mdf_v3.py
@@ -1111,7 +1111,7 @@ class MDF3(object):
             *DataFrame* object. All bytes columns in the pandas *DataFrame*
             must be *latin-1* encoded
         acq_name : str
-            channel group aqcuisition name
+            channel group acquisition name
         acq_source : asammdf.source_utils.Source
             channel group acquisition source
         comment : str
--- a/asammdf/blocks/mdf_v3.py
+++ b/asammdf/blocks/mdf_v3.py
@@ -1090,7 +1090,15 @@ class MDF3(object):
 
             group.trigger = trigger
 
-    def append(self, signals, comment="Python", common_timebase=False, units=None):
+    def append(
+        self,
+        signals,
+        acq_name=None,
+        acq_source=None,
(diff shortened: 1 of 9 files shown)


asammdf/blocks/mdf_v3.py
```diff
@@ -1090,9 +1090,7 @@ def add_trigger(self, group, timestamp, pre_time=0, post_time=0, comment=""):
 
             group.trigger = trigger
 
-    def append(
-        self, signals, acquisition_info="Python", common_timebase=False, units=None
-    ):
+    def append(self, signals, comment="Python", common_timebase=False, units=None):
```
danielhrisca, line 1093: I would like to have the same method signature for the `MDF3` and `MDF4` classes

danielhrisca (changes requested):

asammdf/blocks/mdf_v3.py
JulienGrv: Ok then comment is the only argument we can keep since MDF version 3 and 2 don't have acq_name and acq_source. We can have **kwargs for support of acq_name and acq_source:

```python
def append(self, signals, comment="Python", common_timebase=False, units=None, **kwargs):
```

or alternatively, acq_name and acq_source cannot be pass in append() and must be set using attributes setters.

danielhrisca: I think adding dummy arguments to the  mdf3 class would be more explicit

JulienGrv: Oh you're right, I didn't think about this possibility! I will implement that. Thanks

JulienGrv: It's done. I don't know why the diff shows mdf_v3.py as completely different; just select "ignore whitespaces" and it's back to normal.

danielhrisca (approved):

danielhrisca:
Thank you Julien!

JulienGrv:
You're welcome! By the way, you should do a breaking change release since it changes the API entry for append().

PR #446, opened 2020-12-02
add support for source name and source path in mdf.whereis

JulienGrv:


danielhrisca (changes requested):
I suggest using `_get_source_name` and `_get_source_path` directly in `whereis`

```python
        occurences = tuple(
            (gp_idx, cn_idx)
            for gp_idx, cn_idx in occurences
            if (
                source_name is None
                or source_name in self._get_source_name(gp_idx, cn_idx)
            )
            and (
                source_path is None
                or source_name in self._get_source_path(gp_idx, cn_idx)
            )
        )
        return occurences
```

danielhrisca:
Thank you Julien!

PR #453, opened 2020-12-17
refactor setup.py

JulienGrv:
I still don't like that the setup.py has two setup calls. I think it is error-prone so I'd like to see if this refactoring works or not


danielhrisca:
Building the C extension is problematic especially for Windows so in case the build fails we just install the package and use the pure python fallback functions

JulienGrv:
Hi @danielhrisca! :)

What bothers me is that I don't want to update things in 2 places whenever I touch the setup.py. That's why I tried to just isolate the ext_modules options and not the whole setup call. Let's see if it works like this

JulienGrv:
does not seem to work for me locally. Let me see if I can find why

JulienGrv:
Interestingly, if I use `pip install .`, it does not work, but if I use `python setup.py install` it works

danielhrisca:
_pip_ works in mysterious ways

JulienGrv:
Actually, it seems to work locally when I do as follow:

```bash
python setup.py bdist_wheel
pip install dist/asammdf-6.0.2.dev2-cp39-cp39-win_amd64.whl
```

How can I be sure it will work on PyPi?

danielhrisca:
You have the proper setup to compile the binary wheel. This wheel can now be installed on any computer running windows and python 3.9 x64
This still will not help if you need to install from source

JulienGrv:
What do you mean by installing from source?

danielhrisca:
I mean if there is no wheel available to install from, so pip used the ``.tar.gz`` file

JulienGrv:
I couldn't get _pip_ to build the C extensions. Only _setuptools_ through the _setup.py_ does it. I tried with the _setup.py_ from the development branch and it didn't work either.

I read a lot of information online but found nothing helpful so far. It makes me wonder if that is even possible but I don't see reason why it should not be... I will continue investigating this tomorrow and this weekend.

JulienGrv:
But to me, since it was not working with the development branch either, I consider this PR already valid for a merge but I will still investigate a bit more just in case I find any solution so don't merge yet.

JulienGrv:
Ok I found something! If I delete the _pyproject.toml_, then it works again. So as I understand, when the _pyproject.toml_ is present, _pip_ ignores the _setup.py_ and uses the new PEP 517 way of creating wheel, which uses the _setup.cfg_. I will see if I can tell _pip_ to use the _setup.py_ even if _pyproject.toml_ is present. I think it finally starts to become clearer.

JulienGrv:
Hi Daniel, Merry Christmas :)

So, I found out why pip was not compiling the C modules. Numpy was missing in the setup requirements in pyproject.toml. With the new PEP 517, pip creates an isolated environment with the setup requirements, that is used to run the setup.py and numpy was missing from it hence the C modules were not compiled.

Now that numpy is in the setup requirements, it ensures that the C modules will always be built and available. I actually starts to like the new PEP 517 ^^

Do not merge yet, there is something left I'd like to do. I'll tell you when it is ready to merge.

JulienGrv:
Ready to be merged :)

danielhrisca:
Hello Julien, it looks like there are some missing requirements for the testing stage https://travis-ci.org/github/danielhrisca/asammdf/builds/751594776

JulienGrv:
Hi Daniel, I was not aware of Travis CI. I will fix that asap

JulienGrv:
All the lights are green, ready to be merged 👌

danielhrisca:
Hello Julien,

I've made changes to use pytest in travis, please resolve the conflicts. I also suggest adding the cli argument ``--color=yes`` to the [pytest invocations](https://github.com/JulienGrv/asammdf/blob/feature/refactor-setup-script/appveyor.yml#L61)

Have a great year 2021!

danielhrisca:
Thank you for the PR!

PR #473, opened 2021-01-26
terminate text strings at \0

kkarsko:
The current get_text_v4 implementation strips '\0' (and whitespace) from the retrieved data block, but does not terminate the string at the first '\0' if it contains printable text after.

I ran into this when MDF.get() could not find some signals. Turns out their names were stored as b"Original_Name\0\0\0Other_Junk". The other junk was left there by a signal renaming utility.

### DIFF
--- a/asammdf/blocks/utils.py
+++ b/asammdf/blocks/utils.py
@@ -244,13 +244,13 @@ def get_text_v4(address, stream, mapped=False, decode=True):
         block_id, size = BLK_COMMON_uf(stream, address)
         if block_id not in (b"##TX", b"##MD"):
             return "" if decode else b""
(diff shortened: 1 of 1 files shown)


danielhrisca:
Hello @kkarsko  and thanks for the PR

I think the following code will perform better
```python
text_bytes = stream[address + 24 : address + size].split(b"\0")[0].rstrip(b" \r\t\n\0")
```

And please also update the `get_text_v3` function

danielhrisca:
Do you know which utility was used for renaming?

kkarsko:
It's a proprietary tool that obfuscates signal names to protect IP

danielhrisca:
It doesn't do a very good job 😄

danielhrisca (approved):

danielhrisca:
Thank you @kkarsko!

PR #482, opened 2021-02-03
Update mdf_common.py

Eisdrache:
typing error leads to 
AttributeError: 'MDF4' object has no attribute '_raise_on_mutiple_occurences'
just one 'l' missing 
→ self._raise_on_multiple_occurences         instead of
→ self._raise_on_mutiple_occurences


danielhrisca:
Thank you @Eisdrache !

PR #494, opened 2021-03-03
add acq_name argument to whereis

JulienGrv:
I made this following a discussion I had today with gitMesa. See #493


danielhrisca:
Thanks Julien!

PR #495, opened 2021-03-04
Follow the abstract class design pattern

JulienGrv:
Hi @danielhrisca 

If I am correct, MDF_Common is an abstract class so I propose this refactoring for _filter_occurrences which better follows this design pattern

### DIFF
--- a/asammdf/blocks/mdf_common.py
+++ b/asammdf/blocks/mdf_common.py
@@ -2,6 +2,7 @@
 ASAM MDF version 4 file format module
 """
 
+from abc import ABC, abstractmethod
 import logging
 
 from .utils import MdfException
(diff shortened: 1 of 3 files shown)


danielhrisca:
What I had in mind with the class was to use it as a mixin and avoid code duplication

JulienGrv:
Ok. I will let it as a Mixin but I prefer to make two _filter_occurrences implementations otherwise I have to do this `if self.version >= "4.0"`

danielhrisca:
> Ok. I will let it as a Mixin but I prefer to make two _filter_occurrences implementations otherwise I have to do this `if self.version >= "4.00"`

ok, sounds good

danielhrisca:
Thank you Julien!

PR #498, opened 2021-03-22
Fix segmentation violation when cutting files

crogre:
This includes a few issues found in the extract() function:

1. Align types with the Python interface (signed vs unsigned)
2. Calculate the size of the buffer coherently (sign issue again)
3. Avoid pointer arithmetic


danielhrisca:
Thank you @crogre !

PR #541, opened 2021-07-20
Fix AttributeError when clicking plot

jvanthoog:
Fixes error when clicking or right-clicking the plot: `<class 'AttributeError'>: module 'PyQt5.QtGui' has no attribute 'QApplication'`

All other instances of QApplication in the code use `QWidgets.QApplication`


danielhrisca:
Thank you for the fix!

PR #544, opened 2021-07-23
Fix drag behavior of TreeWidget

jvanthoog:
TreeWidget was using `mouseMoveEvent` to detect a drag, which meant if you moved your mouse even one pixel while clicking an item, you'd start a drag instead (I kept accidentally doing this). I replaced this with `startDrag`, which has a threshold before starting a drag.


danielhrisca:
Thank you @jvanthoog !

PR #548, opened 2021-07-27
Use Pattern.search instead of match for search patterns

jvanthoog:
This replaces `Pattern.match` with `search` for search dialogs. `match` only checks the beginning of the string, while `search` checks anywhere in the string. This means you no longer need to surround search terms with `*` to search for a word anywhere in a channel name.

I'm not sure what was the original reason for using match, but IMO this is more natural (maybe you'll disagree). It matches how a search bar usually works, and is simpler for the more typical case of searching for a word anywhere in a name. You can restore the old behavior in Regex mode with the caret `^`.


danielhrisca:
I think the main reason was that with large channel count `search` performs worse than `match`

danielhrisca:
@jvanthoog please make the PR on the development branch and I will merge it

PR #556, opened 2021-08-12
Prevent freezing during long tasks

jvanthoog:
This makes sure the application continues to process events while a progress bar is showing. Otherwise it can appear to freeze or not update progress.


danielhrisca:
Thank you @jvanthoog

PR #587, opened 2021-10-17
style: apply black and isort

JulienGrv:
This will cleanup PR diffs


danielhrisca:
Did you run isort and black with default settings?

JulienGrv:
Yes

PR #597, opened 2021-11-03
[MDF] Use float_interpolation_mode in to_dataframe call

### LINKED ISSUE #596: [Bug] mdf.to_dataframe() does not respect float_interpolation_mode
## Python version
3.7

## asammdf version: 
6.3.2

  ## MDF version
4.10

  ## Code snippet
```
mdf2.configure(float_interpolation = 0)
df = mdf.to_dataframe(raster = 0.1)
```
## Description
This line: 
https://github.com/danielhrisca/asammdf/blob/06ff7bf672520aa7e1b08325b3cd291cae0eb139/asammdf/mdf.py#L3847

Should be changed to
```
                    signal.interp(
                        master,
                        integer_interpolation_mode = self._integer_interpolation,
                        float_interpolation_mode = self._float_interpolation
                    )
```

Currently `to_dataframe` always uses float interpolation mode 1, but it's essential for my use case that we don't attempt to interpolate float values in the dataframe.

AppliedDev:
Fixes `interpolation_mode` in the `to_dataframe` call. Previously float interpolation always used the default mode 1.

closes #596

### DIFF
--- a/asammdf/mdf.py
+++ b/asammdf/mdf.py
@@ -3844,7 +3844,11 @@ class MDF:
                 cycles = len(group_master)
 
                 signals = [
-                    signal.interp(master, self._integer_interpolation)
+                    signal.interp(
+                        master,
+                        integer_interpolation_mode = self._integer_interpolation,
+                        float_interpolation_mode = self._float_interpolation
+                    )
                     if not same_master or len(signal) != cycles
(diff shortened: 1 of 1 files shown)


danielhrisca:
Thank you for the PR!

PR #600, opened 2021-11-05
fix(blocks.v2_v3_blocks): ModuleNotFoundError: No module named 'pwd'

JulienGrv:


danielhrisca:
`getuser`  seems to work on Windows. On which OS did you have problems?

JulienGrv:
on Windows, the environment variables are all empty in my case, so it tries to fallback on using pwd but pwd is no available on Windows so it raises ModuleNotFoundError

JulienGrv:
Have a look here https://github.com/JulienGrv/asammdf/runs/4118929891?check_suite_focus=true#step:5:355

danielhrisca:
I had no idea this does not work. Thanks!

PR #601, opened 2021-11-05
Streamline continuous integration between remote and local

JulienGrv:
Hi @danielhrisca 

Thanks to tox, we can run tests locally just like it will be on the remote.
Currently, it runs unit tests against python version 3.7, 3.8, 3.9 and 3.10 on ubuntu, windows and macos. It does a style check with black and isort and generates the documentation.

Can you tell me what is missing so that we can completely drop appveyor and travis?


danielhrisca:
The `appveyor` jobs that are run on the `master` branch generate the standalone executable s (using PyInstaller).

I think that the documentation is generated automatically by `read the docs`.

I think that `travis` could be disabled.

JulienGrv:
There are 3 workflows since they are trigger on different events:

- continuous integration (main.yml): this one is triggered on any push, it runs all the tests on every platform / python version.
- coveralls: this one is only triggered for pushes and pull request on the master or development branch.
- build: triggered for push that contains a tag that follows semantic versioning (example: 3.4.5). It creates wheels for all platform / python version. It also create one executable for each platform.

Ok so I think that with this, we can drop appveyor. What do you think?

danielhrisca:
This is some heavy lifting work that you did.

If the wheels and executables can be generated with the new workflow then we can drop appveyor

JulienGrv:
Wheels and executables can be generated. You can actually test by yourself, if you checkout the branch and then create a tag like 3.6.7 and push it.

JulienGrv:
Honestly I did this heavy lifting because I wanted to be able to locally reproduce the tests happening remotely and I thought it's a good exercise to learn github actions. Feel free to play with my branch and test if it does what you need

JulienGrv:
Currently the branch does not remove appveyor so we can merge it as is and you cna start using github actions and when you feel confident enough you can drop appveyor yourself

JulienGrv:
Also I just realized that appveyor is broken currently so this could save the day :)

danielhrisca:
Very nice work Julien!

danielhrisca:
I've pushed a tag but I can't find the exe builds. Is this only valid for the `master` branch?

JulienGrv:
It should work for any branch

JulienGrv:
Ah I see why, we have to adapt the pattern to support dev releases, look here

PR #604, opened 2021-11-09
feat(ci): build artifacts on dev releases

JulienGrv:


JulienGrv:
@danielhrisca just to let you know, there is the possibility to automatically publish the wheels and tarball to PyPi but I could not implement it since I am not the owner. See [here](https://github.com/danielhrisca/asammdf/blob/047c35e2b81279b52772304986658760b2810437/.github/workflows/build.yml#L45)

PR #611, opened 2021-11-22
fix some incorrect type annotations

JulienGrv:


danielhrisca:
thank you Julien!

danielhrisca:
BTW if you use the Tabular window you should check out the development branch since I've switched to a new implementation

danielhrisca:
@JulienGrv do you have further ideas for the 7.0.0 release?

JulienGrv:
No I think it's good now, I've checked the type annotations multiple times now

JulienGrv:
And I have no other breaking change in mind

PR #631, opened 2022-01-06
Stops reading from out side of file bounds.

david32:
Encountered when reading empty files which do not end correctly, the program tried to read beyond file limits.
I encountered it with trying to read a channel. 

Although it is a case of an erroneous mf4 file, still such thing shouldn't happen.


danielhrisca:
Thank you @david32 for this fix!

PR #633, opened 2022-01-07
Added characters in file name to be replaced

david32:
Characters which are illegal (in Windows) for file names can be present in channel comments, escape characters could also be used in comments. 
When exporting to CSV without "single time base" the channel specific csv name contains the comment.

List of illegal characters on Windows: 
https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file


danielhrisca:
LGTM

Issue #640, opened 2022-02-02
Triger comments always added and can't be deleted


jesberpen:
# Python version
_Please run the following snippet and write the output here_
('python=3.10.2 (tags/v3.10.2:a58ebcc, Jan 17 2022, 14:12:15) [MSC v.1929 64 '
 'bit (AMD64)]')
'os=Windows-10-10.0.19044-SP0'
'numpy=1.22.1'
'asammdf=7.0.4'

```python
import platform
import sys
from pprint import pprint

pprint("python=" + sys.version)
pprint("os=" + platform.platform())

try:
    import numpy
    pprint("numpy=" + numpy.__version__)
except ImportError:
    pass

try:
    import asammdf
    pprint("asammdf=" + asammdf.__version__)
except ImportError:
    pass
```
# Code 

  ## MDF version
  _please write here the file version (you can run ``print(MDF(file).version)``)

  ## Code snippet
  _please write here the code snippet that triggers the error_
  
  ## Traceback
  _please write here the error traceback_
  
# Description

The fastest way to debug is to have the original file. For data protection you can use the static 
method _scramble_ to scramble all text blocks, and send the scrambled file by e-mail.

_Please describe the issue here._

Hello,
Using the GUI version of asammdf and the trigger comment functionality (Ctrl+I) in the plots. No matter if I chose accept or cancel it always adds one to the plot. Aside from that I can't find a way to remove the trigger comments, is it possible to do?
Thanks!

danielhrisca:
The triggers are not added now if you press the Cancel button (development branch code). Deleting triggers is not possible

PR #647, opened 2022-02-06
Do not add comment if pressed Cancel

david32:
Issue #640 

Do not add comments if user pressed cancel in the add window comment.


danielhrisca:
Thank you!

Issue #646, opened 2022-02-04
Missing comment xml extraction in File Info Tab


david32:
# Python version
('python=3.10.0 (tags/v3.10.0:b494f59, Oct  4 2021, 19:00:18) [MSC v.1929 64 '
 'bit (AMD64)]')
'os=Windows-10-10.0.19044-SP0'
'numpy=1.22.0'
'asammdf=7.0.4.dev5'  (yes I know there is a newer version but the issue is there as well.)

# Description

Comment extraction missing in File Info tab, exists in all other tabs

In file Info tab [gui/widget/file.py:~403]
```
for i, group in enumerate(self.mdf.groups):
                channel_group = group.channel_group
                if hasattr(channel_group, "comment"):
                    comment = channel_group.comment
                else:
                    comment = ""
```

should be:

```
for i, group in enumerate(self.mdf.groups):
                channel_group = group.channel_group
                if hasattr(channel_group, "comment"):
                    comment = channel_group.comment
                    comment = extract_cncomment_xml(comment)
                    
                    # or perhaps
                    #  comment = extract_cncomment_xml(channel_group.comment)
                    
                else:
                    comment = ""
```

Lets agree on a version, and I will pull request.

danielhrisca:
Either way works for me

david32:
Fixed in the merged pull request

PR #648, opened 2022-02-07
missing XML comment extraction and style

david32:
Issue #646

### DIFF
--- a/asammdf/gui/widgets/file.py
+++ b/asammdf/gui/widgets/file.py
@@ -403,7 +403,7 @@ class FileWidget(WithMDIArea, Ui_file_widget, QtWidgets.QWidget):
             for i, group in enumerate(self.mdf.groups):
                 channel_group = group.channel_group
                 if hasattr(channel_group, "comment"):
-                    comment = channel_group.comment
+                    comment = extract_cncomment_xml(channel_group.comment)
                 else:
                     comment = ""
                 if comment:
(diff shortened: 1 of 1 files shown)


asammdf/gui/widgets/file.py
```diff
@@ -465,7 +465,7 @@ def __init__(
 
                 item = QtWidgets.QTreeWidgetItem()
                 item.setText(0, "Raw size")
-                item.setText(1, f"{size / 1024 / 1024:.1f} MB")
+                    item.setText(1, f"{size / 1024 / 1024:.1f} MB")
```
danielhrisca, line 468: I believe that this indentation is wrong here

danielhrisca (changes requested):

asammdf/gui/widgets/file.py
david32: Indeed, its a remainder from a local change I made for the channel size as discussed in #645 , my bad.

asammdf/gui/widgets/file.py
```diff
@@ -680,8 +680,7 @@ def _update_channel_tree(self, index=None, widget=None):
             for i, group in enumerate(self.mdf.groups):
                 entry = i, 0xFFFFFFFFFFFFFFFF
                 channel_group = TreeItem(entry, mdf_uuid=self.uuid)
-                comment = group.channel_group.comment
-                comment = extract_cncomment_xml(comment)
+                comment = extract_cncomment_xml(channel_group.comment)
```
david32, line 683: correct line is:
comment = extract_cncomment_xml(group.channel_group.comment)

PR #674, opened 2022-03-24
Pointer out of range when reading corrupted file

david32:
Encountered with a corrupted file. 
There is a python exception when the address in the corrupted file is way outside the file size.

added mdf exception to stop reading the file. 

One issue is with mmap, it doesn't contain the name of the file that its reading, so the exception message doesn't contain the name of the file.

+ Automatic Black formatting


danielhrisca:
Thank you David

PR #678, opened 2022-03-28
Fix a bug that accidently overwrites hd comments common properties

SoerenKa93:
Fix a bug that accidently overwrites hd comments common properties e tags with wrong informations
-> project and subject becoming author if they get passed in an own xml structure
bugreport
.


danielhrisca:
Thank you @SoerenKa93 !

PR #682, opened 2022-04-03
Display Channel Group name in channel list

david32:
Related to #645 

Added a setting button to write the channel group name in the channel list.
Changing this setting updates both "Channel" and "Modify && Export" Tab. 
The file info tab is affected only when opening a file.

As I don't have any files that don't have anything in channel_group.acq_name, I couldn't test this case. Perhaps a check for this variable needs to be done before assigning it?


asammdf/gui/widgets/file.py
```diff
@@ -639,15 +647,18 @@ def set_raster_type(self, event):
...
+            self._update_channel_tree(widget=widget)
+
     def _update_channel_tree(self, index=None, widget=None):
 
         if widget is None:
             widget = self.channels_tree
         if widget is self.channels_tree and self.channel_view.currentIndex() == -1:
             return
-        elif widget is self.filter_tree and (
-            self.filter_view.currentIndex() == -1 or not self._show_filter_tree
```
danielhrisca, line 649: the idea here was to have lazy loading of the filter channel list since it is not used as much

asammdf/gui/widgets/file.py
```diff
@@ -2014,59 +2034,7 @@ def aspect_changed(self, index):
                 self.raster_channel.addItems(self.channels_db_items)
                 self.raster_channel.setMinimumWidth(100)
 
-            if not self._show_filter_tree:
```
danielhrisca, line 2017: again lazy loading is necessary here

PR #689, opened 2022-04-26
Allow iteration over mdf file object to return raw signals

david32:
When iterating over an MDF file, a channel object is returned.
By default this is a converted channel. So the iteration takes a long time because every channel is converted.
Moreover, there is no possibility to convert a channel back to raw, only from raw to converted.

So I thought adding the option to iterate over channels without converting them would be nice and faster.
The default behavior remains the same, although after looking at the performance difference, I would say that perhaps the default behavior should be iterating raw signals.

The results of the tests I've made: 

```
Iteration speed test over file of size: 103.8 MB
 Regular time to iterate over file: 1.91225 seconds
 Raw time to iterate over file: 0.37859 seconds
 Raw + physical convert time to iterate over file: 1.899 seconds

Iteration speed test over file of size: 510.35 MB
 Regular time to iterate over file: 9.54934 seconds
 Raw time to iterate over file: 1.80888 seconds
 Raw + physical convert time to iterate over file: 9.5655 seconds

Iteration speed test over file of size: 811.18 MB
 Regular time to iterate over file: 14.88366 seconds
 Raw time to iterate over file: 2.72116 seconds
 Raw + physical convert time to iterate over file: 15.32696 seconds
```

### DIFF
--- a/asammdf/mdf.py
+++ b/asammdf/mdf.py
@@ -569,8 +569,8 @@ class MDF:
                         ev_type = v4c.EVENT_TYPE_START_RECORDING_TRIGGER
                     event = EventBlock(
                         event_type=ev_type,
-                        sync_base=int(timestamp * 10**9),
-                        sync_factor=10**-9,
+                        sync_base=int(timestamp * 10 ** 9),
+                        sync_factor=10 ** -9,
                         scope_0_addr=0,
                     )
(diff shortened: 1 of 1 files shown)


danielhrisca:
Looks ok thank you!

PR #664, opened 2022-03-01
Reordering Channel Groups when concatenating

david32:
Concerning Issue #663 

Tested with the files I have (mdf 4.1).

PR #690, opened 2022-04-26
Reordering Channel Groups when concatenating v2

david32:
Continuing PR #664 
Apparently I messed up that PR, which was a surprise to me. (Sorry for the mess)
In any case, I found that an improvement is needed as some groups can have a similar name, path, etc. but different channel names. 
As such I've also added a test for names of channels.
Should work better now. 

I tested it by concatenating a few few files which have a certain counter, and plotting this counter, which seems to work fine value wise. 
There are issues with the new plotting system, but that is for another PR or issue.


danielhrisca:
Let me know what issue you found related to the Plot

Issue #712, opened 2022-05-31
TypeError on resample: MF4 opened using the channels option


novack96:
# Python version
_Please run the following snippet and write the output here_
```
python=3.8.0 (default, Nov  6 2019, 16:00:02) [MSC v.1916 64 bit (AMD64)]
os=Windows-10-10.0.19041-SP0
numpy=1.22.2
asammdf=7.0.4
```

# Code 

  ## MDF version
  4.10

  ## Code snippet
  ```
import asammdf as mdf
channelFilterList=["A1","A2", "iconInfo"]
etalonSignalName = "iconInfo"
filepath="C:\\meas\\test.mf4"
meas = mdf.MDF(filepath, memory="minimum", channels=channelFilterList)
meas.configure(integer_interpolation=2)
resampled = meas.resample(etalonSignalName)
  ```
  
  ## Traceback
   File "C:\git\kpitool\Preprocess\Preprocess.py", line 149, in resampleSignalsInMDF
    resampled = meas.resample(etalonSignalName)
  File "C:\Users\NOA1BP\.conda\envs\kpitool\lib\site-packages\asammdf\mdf.py", line 2835, in resample
    MDF._transfer_channel_group_data(mdf.groups[dg_cntr].channel_group, cg)
2022-05-31 15:33:20,983 [CRITICAL] TypeError: list indices must be integers or slices, not NoneType
# Description
I have to open and work with large measurement files (~10GB, and with over 1,000,000 signals) so I need to filter and resample the measurement files before the actual data processing, and I ran into this issue. 

I've tried resampling by removing the `channels` and `memory` options from the opening, and that's working just fine, but it takes way too long obviously because there are over a million signals. So I've applied a filter to it, (see below) so at least it is running without error.

 `meas = mdf.MDF(filepath).filter(["A1","A2", "iconInfo"])`

But I think the resampling should be working for the channel filtered MDF object as well.

I cannot send these measurement files due to confidelity aggreements.

Thank you for your time and consideration.

novack96:
Quick update:

I updated to the latest version v7.0.7, and retried but got the same error.

danielhrisca:
The ``memory`` argument has been dropped since 5.0.0

Regarding the error I will have a look later

david32:
It's just trying to resample channels that weren't loaded, see fix in the PR.

novack96:
I've tried the development branch and now it's working flawlessly! 👏 💯 👍

PR #714, opened 2022-05-31
Issue #712

david32:
#712 

empty list was created because those channels weren't loaded.
Continue loop and perform action only to the group of channels that was loaded

+

Black formatting

### DIFF
--- a/asammdf/mdf.py
+++ b/asammdf/mdf.py
@@ -2998,7 +2998,7 @@ class MDF:
                 for ch_index in channel_indexes
             ]
 
-            if len(channels) == 0:
+            if not channels:
                 continue
 
             sigs = self.select(channels, raw=True)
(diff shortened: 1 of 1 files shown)


asammdf/mdf.py
```diff
@@ -2997,6 +2997,10 @@ def resample(
                 ].items()
                 for ch_index in channel_indexes
             ]
+
+            if len(channels) == 0:
```
danielhrisca, line 3001:
```python
if not channels:
    continue
```

is cleaner

danielhrisca (changes requested):

asammdf/mdf.py
david32: Done. :)

danielhrisca:
Thank you!

PR #715, opened 2022-05-31
Dynamically set the width of the channel list in plot

david32:
With the latest changes the name and unit fields width when plotting was very large and took too much space from the plot.

This proposed change makes the field widths change depending on the channel name and unit length.

Ideally it would be possible to either remove some buttons to make the plot area even larger, have it as an option to exclude some of the buttons (such as font increase/decrease) or move them to a second line / buttom of the area. But that should be discussed.

### DIFF
--- a/asammdf/gui/widgets/plot.py
+++ b/asammdf/gui/widgets/plot.py
@@ -1867,48 +1867,28 @@ class Plot(QtWidgets.QWidget):
 
         nameColumnWidth = max(
             [
-                (len(channel.name) + 5) * self.font().pointSize()
-                for channel in channels.values()
+                self.channel_selection.fontMetrics().boundingRect(f"{channel.name}AAAA").width() + 35 # AAAA extra characters
+                for channel in (list(channels.values()) + self.plot.signals)
             ]
         )
-        if (
-            nameColumnWidth
-            > self.channel_selection.columnWidth(self.channel_selection.NameColumn)
-            and nameColumnWidth < 35 * self.font().pointSize()
-        ):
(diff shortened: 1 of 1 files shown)


danielhrisca:
Maybe you should consider refining this unused method instead https://github.com/danielhrisca/asammdf/blob/6075609e895e2b7cfe17b452cd8383603700507a/asammdf/gui/widgets/plot.py#L2034

danielhrisca:
Maybe you should consider refining this unused method instead https://github.com/danielhrisca/asammdf/blob/6075609e895e2b7cfe17b452cd8383603700507a/asammdf/gui/widgets/plot.py#L2034

david32:
Made the changes to reuse adjust_splitter.

Had to use AAAA for padding for .boundingRect() which isn't nice, but necessary 
Also perhaps there is a better way to count the length of a checkbox.
boundingRect() didn't return the values I've expected.

david32:
@danielhrisca 

Do you have any other suggestions or requests for this PR? A different way for padding?

danielhrisca (approved):

danielhrisca:
Merged for now, but maybe we will need adjustments later on

danielhrisca:
@david32 the performance was deeply impacted for plots with large channel count. I've pushed changes to the development branch. Please check the current behavior

david32:
of course your solution is much more elegant. I didn't know such an options existed and didn't search for it - my bad.
I am surprised that there is such a performance impacts, but I haven't test it with more that 10 channels.

danielhrisca:
16s wasted from ~4000 channels

PR #716, opened 2022-05-31
Tests fail when comparing float arrays

david32:
When comparing floats, its more correct to use np.allclose()

Also possible to use numpy testing...


danielhrisca:
Thank you!

PR #719, opened 2022-06-06
Black formatting

david32:
Sorry, I had an older version of black installed leading to wrong formatting that fails the github action style tests.


david32:
Now all tests are failing because of docs in the py3.7 environment. It does pass on my pc with the same Sphinx version...

PR #742, opened 2022-08-01
fix: :bug: CN block within CA block is not implemented yet

JulienGrv:
@danielhrisca could you do a release for this fix? Thanks!


danielhrisca:
Hello Julien, is this like an array of structs?

JulienGrv:
Yes I suppose. It is not implemented yet but I'd like to be able to open the file though so I made this

PR #766, opened 2022-09-13
convert(): removed sharp clips in string parameter

woodpy:
Calling the convert function with "<...>" inside the message parameter let to an invalid XML Error in the Vector MDF Validator inside the File History Meta Data Block


danielhrisca:
LGTM thank you!

Issue #770, opened 2022-09-26
Converting to dataframe converts values to byte strings


eblis:
# Python version
```
('python=3.9.12 (tags/v3.9.12:b28265d, Mar 23 2022, 23:52:46) [MSC v.1929 64 '
 'bit (AMD64)]')
'os=Windows-10-10.0.19044-SP0'
'numpy=1.23.3'
'asammdf=7.1.1'

```
# Code 

  ## MDF version
4.00

  ## Code snippet
I have an MF4 file and when I convert it to dataframe using the following code
```
mdf_file.to_dataframe(channels, raw=True,
                            time_as_date=True,
                            empty_channels="zeros",
                            reduce_memory_usage=True,
                            ignore_value2text_conversions=False)
```
i get the following result:
![image](https://user-images.githubusercontent.com/540188/192274933-37f8b050-8b28-4a80-8d8d-015004e1eb81.png)
The values are normal integers.

However, when I switch to `raw=False`, as such:
```
mdf_file.to_dataframe(channels, raw=False,
                            time_as_date=True,
                            empty_channels="zeros",
                            reduce_memory_usage=True,
                            ignore_value2text_conversions=False)
```
I get the following result:
![image](https://user-images.githubusercontent.com/540188/192275350-3d1ff2eb-ad67-4d41-93d8-96ed17c0e51d.png)
  
  ## Traceback
There's no traceback
  
# Description
I'm not sure I can send the data file, even if scrambled.
The problem seems to be the conversion from raw values to interpreted values. Somewhere along the line the values are converted from integers to byte strings that actually hold a float value (like maybe an extra `str()` was applied or something).

I can go over all values and attempt to convert them to float (not all of them may be actual floats, some values may actually be real strings) but that takes a **looong** time.

Debugging this a bit i can see that the signals that this happens for have their `conversion_type` configured for `7 - v4c.CONVERSION_TYPE_TABX`. 
For example this one:
![image](https://user-images.githubusercontent.com/540188/192278588-b0ee2e60-bfee-4681-a840-bef1ba9889cf.png)
![image](https://user-images.githubusercontent.com/540188/192279105-6b53a936-458d-4c9c-8228-21f298a44209.png)

I believe this was "working" in asammdf 5.x. I'm not sure working is the correct term, as it was probably a bug in that version, but the 5.x version had floats for the actual floats and strings for the rest. Version 7 has strings for everything, but the strings may actually be a float value.

eblis:
[...]

PR #771, opened 2022-09-27
Fix for issue #770. Try to convert to float first, and if that fails convert to bytes

eblis:
Alternatively I see that another method is used further up in the codebase, starting on line 3400:
```
                all_bytes = True
                for v in ret.tolist():
                    if not isinstance(v, bytes):
                        all_bytes = False
                        break

                if not all_bytes:
                    try:
                        ret = ret.astype("f8")
                    except:
                        if not as_object:
                            ret = np.array(
                                [np.nan if isinstance(v, bytes) else v for v in ret]
                            )

                else:
                    ret = ret.astype(bytes)
```
We could use this check to see if we need to convert to `float` or `bytes`, but I wouldn't iterate over all items if we can help it.

### DIFF
--- a/asammdf/blocks/v4_blocks.py
+++ b/asammdf/blocks/v4_blocks.py
@@ -3425,7 +3425,6 @@ class ChannelConversion(_ChannelConversionBase):
                     ],
                 )
             else:
-
                 ret = np.full(values.size, None, "O")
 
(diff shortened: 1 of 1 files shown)


danielhrisca:
thank you @eblis

Issue #741, opened 2022-07-29
Progress callback is called too soon ?


eblis:
# Python version
```
('python=3.9.12 (tags/v3.9.12:b28265d, Mar 23 2022, 23:52:46) [MSC v.1929 64 '
 'bit (AMD64)]')
'os=Windows-10-10.0.19044-SP0'
'numpy=1.23.1'
ldf is not supported
xls is not supported
xlsx is not supported
'asammdf=7.1.0'
```
# Code 
  ## Code snippet
```python

logger = logging.get_logger(__name__)

def load_progress(progress_bar: tqdm.tqdm, current, total):
    if progress_bar.total != total:
        progress_bar.clear()
        progress_bar.reset(total=total)
    progress_bar.update(1)
    progress_bar.refresh()

loading = tqdm.tqdm(total=100, leave=False, desc=f"Loading {file.name}")
progress_function = functools.partial(load_progress, loading)
try:
    with MDF(file, callback=progress_function) as mdf_file:
        loading.close()  # done with load progress
        logger.info(f"Reading finished for {file}")
```

# Description
The progress for loading the file finishes a lot faster than the file is actually loaded.
The progress for my file (but the file itself doesn't matter from what I can tell) has total set to `13903` or similar and the progress finishes in a few seconds, but the progress bar is closed and the log message saying `Reading finished for {file}` is only written minutes later

danielhrisca:
``current`` and ``total`` are not confined in the interval 0..100 
Can this be a problem for your progress bar?

eblis:
If the `total` value reported by asammdf is different than the current one (default 100) the progress bar is reset to use the new `total` as the maximum. And every time it's called the progress is increased by `1`.
If you do a `progress_bar.write(f"{current}: {total}")` in the callback function you can see that value goes from 0 to whatever `total` has, but the file hasn't finished loading at that point, it still takes a lot longer after callback has been called for the last time

eblis:
So it turns out that the problem is the sort call after reading the data.

I added some quick print statements to the `_read()` method (for mdf_v4) and this is what I got. I measured the time it took from the start of the method call to the `self._sort()` call which seems to load the data, the time it took for the `self._sort()` call and then the time after sorting until the end, which seems to be handling some kind of events.

```
Reading took 9.03205919265747 seconds
Sorting took 1453.9986987113953 seconds
Events took 12.37953495979309 seconds
```

eblis:
[...]

PR #772, opened 2022-09-27
Fix for issue #741.

eblis:
Added progress information while sorting all the groups.
 - Added 102 extra steps in the load progress, 100 for sorting and 2 after sorting is completed.
 - When incrementing the progress updates during sorting we first compute a step, based on the total number of steps we have (100), the number of groups and the total items and then we increment each sorting by the step count.
 - If the previous step count (integer) is greater than the current step (integer) then we can update the progress. This avoids calling the callback function with floating values and avoids calling the callback function with the same current value of the progress bar - i.e. we only call when there's an update in progress.

### DIFF
--- a/asammdf/blocks/mdf_v4.py
+++ b/asammdf/blocks/mdf_v4.py
@@ -778,9 +778,13 @@ class MDF4(MDF_Common):
                     else:
                         break
 
-        self._sort(current_progress_index=current_cg_index, max_progress_count=progress_steps)
+        self._sort(
+            current_progress_index=current_cg_index, max_progress_count=progress_steps
+        )
         if self._callback:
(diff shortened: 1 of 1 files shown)


danielhrisca:
@eblis please resolve the merging conflict

danielhrisca:
@eblis please try out the development branch code (you need to recompile the c extension code as well) for a big speed-up in the sorting

eblis:
Sorting speed improvement sounds very good.
But still the progress stuff should help, no ? It will still report during the sorting steps, even though they might be a [lot] faster now.

danielhrisca:
thank you!

danielhrisca:
@eblis are you using a custom _callback function?

PR #780, opened 2022-10-25
update requirement for extras [gui]

jazi007:
With [pyqtgraph-0.13.1](https://github.com/pyqtgraph/pyqtgraph/releases/tag/pyqtgraph-0.13.1)
This import https://github.com/danielhrisca/asammdf/blob/development/asammdf/gui/widgets/plot.py#L18 will fails.

I see that some of that update is done for some of changes here: https://github.com/danielhrisca/asammdf/blob/development/asammdf/gui/widgets/plot.py#L23 

So either applying same technic:
```python
try:
    import pyqtgraph.console.template_pyside6
except ImportError:
    import pyqtgraph.console.template_generic
```

or make sure same version of packages used in CI are used in setup.py

let me know if I need to update this PR

PR #797, opened 2022-12-06
gui broken by import

tainnok:
this PR is basicaly a cleaner version of #780 
It's the same problem (import not possible), but this PR does NOT contain any other changes (like #780) and the fix does NOT introduce a new try-except block.
It just moves the import into the existing try-except block.

ATTENTION: on python3.10 PR #796 ("do not limit SciPy version to <1.8.0 for GUI") is needed to run tests without errors!


danielhrisca (approved):

danielhrisca:
Thank you @tainnok

PR #807, opened 2023-01-05
bugfix: do not throw a AttributeError in case of an empty dict (conversion)

tainnok:
method: conversion_transfer()
it seems that in some cases the argument 'conversion' is an empty dictionary.

an empty dict need to be handled the same way as None (which was already handled).


danielhrisca:
thank you @tainnok

PR #824, opened 2023-03-02
Used regular expression

KevinRodrigues05:
Hi Daniel,

This is a small bug fix we realized when we could not extract the comment from the mdf file.
There seems to be an issue with the double and single quotes specified in the replace string.

The comment(variable) string seems to have single quotes in some of our mf4 files -` xmlns='http://www.asam.net/mdf/v4'`. 

Hence I have provided a fix which will take care of both the cases (double & single quotes)

It is also tested on different mf4 files with my colleague Monica K. @kmonica92 

Is this solution good enough if it works for you as well.


danielhrisca:
Thank you Kevin!

PR #839, opened 2023-03-26
QtSignal.emit() missing parameter

david32:
When modifying/extracting bus with the GUI, at the end of the process, there is an exception where the self.process QtSignal is missing a parameter.
This is because this signal is connected to a function expecting (self) parameter but none is given
The solution is to pass self.parent to it.

### DIFF
--- a/src/asammdf/gui/utils.py
+++ b/src/asammdf/gui/utils.py
@@ -306,7 +306,7 @@ class ProgressDialog(QtWidgets.QProgressDialog):
         self.error = error
 
     def thread_complete(self):
-        self.finished.emit()
+        self.finished.emit(self.parent)
         self.thread_finished = True
         super().close()
 


danielhrisca:
Thank you David

danielhrisca:
@david32 I had to revert this PR. Please share your initial error message

david32:
image


```
 File "D:\Python_Projects\asammdf\src\asammdf\gui\utils.py", line 309, in thread_complete
    self.finished.emit()

<class 'TypeError'>: finished(int) needs 1 argument(s), 0 given!
```

As mentioned, when you use the connect function for the signal and then emit, you need to pass the arguments to that function, but provide none.

danielhrisca:
I think that was just a refactoring mistake in the development version

david32:
image

This is straight from the executable that I downloaded (checked the same for 7.3.3)

Is it not reproducable on your machine? 
The result of whatever you are trying to do (modify/apply DBCs) is created, so the file is there, its just the GUI that shows this fault.

danielhrisca:
try 7.3.8

david32:
image


Why would there be a difference? is the explanation wrong? do you disagree with it? I don't get it.

danielhrisca:
What confuses me is that there is no argument în the signal definition
https://github.com/danielhrisca/asammdf/blob/d846d3349904f3fcaf33201c3dc4faaa7c67b875/src/asammdf/gui/utils.py#L270

david32:
Yes, I was confused by that as well, but once you connect, it expects an argument:

```
from PySide6 import QtCore
class TestSignalClass(QtCore.QObject):
    somesignal = QtCore.Signal()

    def __init__(self, *args, **kwargs):
        super().__init__( *args, **kwargs)
    
    def close_test(self):
        self.somesignal.emit()
        
def double_me(num : int):
    print(num*2)

def test():
    t = TestSignalClass()    
    t.somesignal.connect(double_me)
    t.close_test()

if __name__ == "__main__":
    test()
```

I also noticed now while playing with this test that if I add an argument to emit() after connect in this case - another exception pops `TypeError: somesignal() only accepts 0 argument(s), 1 given!`
Perhaps it is a bug with QtCore, I don't know, but in any case, the GUI is unusable without this modification at the moment.
(also GUI wise, scrolling on the vertical axis in a plot no longer works, I will perhaps look further into it or will open a ticket)

PR #857, opened 2023-04-27
Update v4_blocks.py

chr-sy:
tz-naive datetimes should now be handled as required by the MDF spec: if the start_time setter function gets a tz-naive datetime, it produces a "local" timestamp which is converted to the exactly same tz-naive datetime by the getter function.

### DIFF
--- a/src/asammdf/blocks/v4_blocks.py
+++ b/src/asammdf/blocks/v4_blocks.py
@@ -5448,13 +5448,17 @@ class FileHistory:
         """
 
         timestamp = self.abs_time / 10**9
+        tz_local = False
         if self.time_flags & v4c.FLAG_HD_LOCAL_TIME:
-            tz = dateutil.tz.tzlocal()
+            tz_local = True
+            tz = timezone.utc
         else:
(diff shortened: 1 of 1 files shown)


src/asammdf/blocks/v4_blocks.py
```diff
@@ -5465,7 +5469,7 @@ def time_stamp(self) -> datetime:
     def time_stamp(self, timestamp: datetime) -> None:
         if timestamp.tzinfo is None:
             self.time_flags = v4c.FLAG_HD_LOCAL_TIME
-            self.abs_time = int(timestamp.timestamp() * 10**9)
+            self.abs_time = int(timestamp.replace(tzinfo=timezone.utc).timestamp() * 10**9)
```
EPanni, line 5472: In addition to making it more readable, the multiplier constant is used a lot so I suggest having a specific object with that value and multiplying by that value where necessary.

Idk, something like:

SECONDS_TO_NANOSSECONDS_MULTIPLIER = 10**9

self.abs_time = int(timestamp.replace(tzinfo=timezone.utc).timestamp() * SECONDS_TO_NANOSSECONDS_MULTIPLIER)

Issue #859, opened 2023-05-08
Reduce_memory_usage and invalidation_bits behaviour


car-efficiency:
# Python version
_Please run the following snippet and write the output here_
'python=3.9.5 (default, Jul  5 2021, 13:54:33) \n[GCC 9.2.0]'
'os=Linux-3.10.0-1160.76.1.el7.x86_64-x86_64-with-glibc2.17'
'numpy=1.23.5'
'asammdf=7.3.12'
# Code 

  ## MDF version
 4.00

  ## Code snippet
```
                        df_mf4_stacked=mf4_stacked.to_dataframe(
                            channels=channels,
                            raster=raster,
                            time_from_zero=True,
                            reduce_memory_usage=True,
                            ignore_value2text_conversions=True,
                        )
```
  
  ## Traceback

>   File "/home/../lib/python3.9/site-packages/asammdf/mdf.py", line 4493, in to_dataframe
>     result = mdf.to_dataframe(
>   File "/home/../lib/python3.9/site-packages/asammdf/mdf.py", line 4736, in to_dataframe
>     sig.samples = downcast(sig.samples)
>   File "/home/../lib/python3.9/site-packages/asammdf/blocks/utils.py", line 1642, in downcast
>     min_ = array.min()
>   File "/home/../lib/python3.9/site-packages/numpy/core/_methods.py", line 44, in _amin
>     return umr_minimum(a, axis, None, out, keepdims, initial, where)
> ValueError: zero-size array to reduction operation minimum which has no identity

  
# Description

The creation of a dataframe fails. One signal in the mf4 file has all invalidation_bits in the array are set to "true". The signal samples are an array of zeros.
![image](https://user-images.githubusercontent.com/132067129/236816443-a5141fb0-5482-4c2c-a223-16325841d748.png)
![image](https://user-images.githubusercontent.com/132067129/236816515-d635cc3c-2d37-49d0-b7e0-862606bcf400.png)

Therefore, the "validate" method in signal.py removes all signal samples and returns a signal object with an empty array.
![image](https://user-images.githubusercontent.com/132067129/236816620-761d444d-65e8-4b05-8663-c486b8b4b0ee.png)

This empty sig.samles array leads to the error in utils.py because we use the reduce_memory_usage feature while using with large mf4 files (that is also why we cannot just deactivate reduce_memory_usage). 
[...]

PR #860, opened 2023-05-10
Only downcast signals if the samples array is not empty

car-efficiency:
Hotfix for issue #859. 
No downcast is required if the sig.samples array is empty (e.g. because all signals are invalid).


danielhrisca (approved):

danielhrisca:
Thanks!

PR #879, opened 2023-07-13
Use PySide6 Scoped Enums

zariiii9003:
- Pin PySide6 version to 6.5
- Use scoped enums


danielhrisca:
Thanks for the work @zariiii9003. 

Some functionality is still dependent of older PySide releases (GPS windows do not work with PySide6>6.2.2), I feel like we need to hold back on committing to PySide6.5 for a while

zariiii9003:
I understand, i pinned PySide6 back to 6.2.3. The scoped Enums should work anyway and save you some work in the future.

zariiii9003:
@danielhrisca That bug from my last commit also exists in your latest release.
Also, i'm not sure why that ubuntu test failed.

zariiii9003:
@danielhrisca The GPS widget issue is fixed.
@tov101 Could you help me with the `test_PushButton_SaveOfflineWindows` test? The drag and drop implementation seems to fail on Windows.

tov101:
@zariiii9003 I'll take a look on that test tomorrow.

tov101:
@zariiii9003 Sorry for late response. I have investigated the problem a little bit (maybe not enough) but I could not find a solution yet.
It seems that `startDrag` execution starts own event loop, and keeps us blocked until mouseMove event is intercepted.
However, in this latest version of PySide, if mouse is moved in other thread, on windows at least, `startDrag` end it's execution and we cannot drop anything anywhere.
I'll keep looking for a solution when I'll find some time.

danielhrisca:
@zariiii9003 any way you could resolve the conflicts? 

I've done some local tests and the GPS window works now with PySide6==6.6.0

zariiii9003:
> @zariiii9003 any way you could resolve the conflicts?
> 
> I've done some local tests and the GPS window works now with PySide6==6.6.0

I could try, but all those drag and drop tests failed, and that probably didn't change

zariiii9003:
@danielhrisca i wrote a script for the replacement. I force pushed a singe commit, where i changed only the enums and nothing else. It's probably incompatible with older PySide6 versions though.

danielhrisca (approved):
Works with PySide6==6.3.1 and PySide6==6.6.0
Thank you!

Issue #893, opened 2023-08-18
Different result with iter_to_dataframe and to_dataframe 


r-xela:
# Python version
_Please run the following snippet and write the output here_
```python
('python=3.11.0 (main, Nov  7 2022, 20:37:44) [Clang 14.0.0 '
 '(clang-1400.0.29.202)]')
'os=macOS-13.4.1-x86_64-i386-64bit'
'numpy=1.25.2'
ldf is not supported
xls is not supported
xlsx is not supported
yaml is not supported
'asammdf=7.3.14'
```
# Code 

  ## MDF version
4.10

  ## Code snippet
```
from asammdf import MDF

# Load mdf
mdf = MDF("example_iter_to_dataframe_bug.scrambled.mf4")

# Configure mdf according to PO instructions
mdf.configure(float_interpolation=0, integer_interpolation=0)


iter_df_list = []

# we set raw=False, but still get raw values in the resulting dataframe here
for df_mdf_iter in mdf.iter_to_dataframe(
    raster=0.05,
    reduce_memory_usage=True,
    ignore_value2text_conversions=True,
    time_from_zero=False,
    raw = False
):
    iter_df_list.append(df_mdf_iter)

# if we use to_dataframe on same mdf we get correct values
to_dataframe = mdf.to_dataframe(
    raster=0.05,
    reduce_memory_usage=True,
    ignore_value2text_conversions=True,
    time_from_zero=False,
    raw = False
    )

# we would except same data for both dataframes
print(to_dataframe['QUTUAIJHDBXVIFDXXULUXRA'].sample)
print(iter_df_list[0]['QUTUAIJHDBXVIFDXXULUXRA'].sample)
```
  
  ## Traceback
None
  
# Description

The fastest way to debug is to have the original file. For data protection you can use the static 
method _scramble_ to scramble all text blocks, and send the scrambled file by e-mail.

Hi, 
if you run the above code snippet, we would expect the data frames to look identical. But they differ, the iter_to_dataframe just contains the raw values and does not run the conversion. We had a quick look at the source code and think that the issue comes from https://github.com/danielhrisca/asammdf/blob/master/src/asammdf/mdf.py#L4231 - we guess that should be `<=` and not `<`. You can use the attached example file

[example_iter_to_dataframe_bug.scrambled.mf4.zip](https://github.com/danielhrisca/asammdf/files/12379137/example_iter_to_dataframe_bug.scrambled.mf4.zip)

jlyda:
[...]

PR #895, opened 2023-08-23
Aligned iter_to_dataframe with raw=False conversion of to_dataframe m…

jlyda:
Fix for issue https://github.com/danielhrisca/asammdf/issues/893

Aligned `iter_to_dataframe` with `to_dataframe` code for conversion with `raw=False`.


danielhrisca:
@jlyda please have a look at the development branch code

jlyda:
Thanks @danielhrisca! But you aligned `to_dataframe` with `iter_to_dataframe` as reference. Regarding #893 sample data, the `to_dataframe` method was fine, but the case differentiation of `text_conversion` here: https://github.com/danielhrisca/asammdf/blob/development/src/asammdf/mdf.py#L4048 causes that with `mdf.version > 4 and conversion.conversion_type==7` the conversion is skipped.
To solve this, my idea was to align `iter_to_dataframe` with `to_dataframe` which was working fine for the issue #893.

PR #911, opened 2023-09-14
Add support for elist in mf4 header comment

brf2cb:
Hello,

I do not know why you decided to convert everything in the header comment to the tree structure. But I would like to add this simple support for the elist so it is not lost in the header.

Previously elist was working in version 7.0.7


danielhrisca:
Thanks for the PR

PR #913, opened 2023-09-20
There is no attribute which is called "types" so it is changed to the available attribute "signal_types".

hamed2dev:


danielhrisca (approved):

danielhrisca:
@hamed2dev do you use the format 4.20? Which application are you using for validation?

hamed2dev:
@danielhrisca Yes I use the format 4.20 for some MF4 files. If you mean how I check the files by validation, I use VSignlyzer to check the signals when it is only looking at the signals. I wanted to stack files with 4.10 format and 4.20 format, where I found out the bug. What I still did not understand is why the version of the last file is taken as the version of the output file of stacking, but since it did not cause any problem I did not change it.  ;)

danielhrisca:
Thanks!

danielhrisca:
>  What I still did not understand is why the version of the last file is taken as the version of the output file of stacking, but since it did not cause any problem I did not change it. ;)

It's simply using the default value for the `version` which is `4.10` (see https://asammdf.readthedocs.io/en/latest/api.html#asammdf.mdf.MDF.stack)

PR #925, opened 2023-10-14
Bug fixes

zariiii9003:
Carry-over of two bug fixes from #879
- ValueError because of read-only array
- fix GPS map not being loaded on newer PySide6 versions due to net::ERR_NETWORK_ACCESS_DENIED


danielhrisca:
Thank you @zariiii9003 !

PR #926, opened 2023-10-15
Use ruff for pyupgrade and isort

zariiii9003:
You could activate many more [rules](https://docs.astral.sh/ruff/rules/) to be checked on commit and in CI.

### DIFF
--- a/src/asammdf/blocks/conversion_utils.py
+++ b/src/asammdf/blocks/conversion_utils.py
@@ -1,4 +1,3 @@
-# -*- coding: utf-8 -*-
 """
 asammdf utility functions for channel conversions
 """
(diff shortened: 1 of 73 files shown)


src/asammdf/blocks/conversion_utils.py
```diff
@@ -379,7 +378,7 @@ def from_dict(conversion: dict[str, Any]) -> v4b.ChannelConversion:
     return conversion
 
 
-def to_dict(conversion: ChannelConversionType) -> Union[dict, None]:
+def to_dict(conversion: ChannelConversionType) -> dict | None:
```
danielhrisca, line 381: Doesn't this require Python >= 3.10 ?

zariiii9003: Yes it does, but ruff ignores it, because the file contains `from __future__ import annotations`.

danielhrisca: Ok, let me know when it is ready for merging

zariiii9003: I'd say it's ready. This little change resulted in 73 modified files, that's enough for one PR 😄

danielhrisca:
Thanks!

PR #942, opened 2023-11-02
unnecessary print statement removed

morbult:
This print statement seems unnecessary. It originates from 4e2b2d27af1ed1312fdb4276bf4a80a9fa8e546e but I cant't see the purpose. Could it just be removed?

For me it started to appear just now that I changed my Python environment. Everytime I run Signal.plot I get an error printed (no actual error, just the traceback printed) because the call to gui.plot 4 lines above the print raises an Exception since importing PySide6 fails in my environment. But Signal.plot produces a plot, so no need to show an error I guess.

### DIFF
--- a/src/asammdf/signal.py
+++ b/src/asammdf/signal.py
@@ -5,7 +5,6 @@ from __future__ import annotations
 from collections.abc import Iterator
 import logging
 from textwrap import fill
-from traceback import format_exc
 from typing import Any
 
 import numpy as np
--- a/src/asammdf/signal.py
+++ b/src/asammdf/signal.py
@@ -224,7 +224,6 @@ class Signal:
             return
 
         except:
-            print(format_exc())
             try:
                 import matplotlib.pyplot as plt
                 from matplotlib.widgets import Slider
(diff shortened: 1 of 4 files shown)


danielhrisca:
the style test is failing, you need to run this file to fix the problems https://github.com/danielhrisca/asammdf/blob/development/run_black_and_ruff.bat

morbult:
Sorry for the delay (and using the master branch) but now I added the changes from running the style checks in development/run_black_and_ruff.bat

morbult:
Not really sure how to proceed here. Style checks fail because of 
```
style: commands[1]> ruff check ./src
  src/asammdf/signal.py:8:23: F401 [*] `traceback.format_exc` imported but unused
```
but ruff doesn't pick up on that when I run locally. Of course I can remove that import manually, but it doesn't have anything to do with my PR.

Regarding the general tests which fails on Ubuntu and Windows in test_PushButton_CreateWindow and test_Plot_ChannelSelection_DragAndDrop_fromPlot_toPlot, I don't see how that is related to my PR either.

morbult:
Style should be fixed now

danielhrisca (approved):

danielhrisca:
Thanks!

PR #976, opened 2024-01-08
Activate Pylint Rules

zariiii9003:
Can you check these?

```text
  style: commands[1]> ruff check ./src
  src/asammdf/gui/widgets/plot.py:136:8: PLR0124 Name compared with itself, consider replacing `b != b`
  src/asammdf/gui/widgets/plot.py:144:8: PLR0124 Name compared with itself, consider replacing `b != b`
  src/asammdf/gui/widgets/plot.py:4460:28: PLR0124 Name compared with itself, consider replacing `min_ != min_`
  src/asammdf/gui/widgets/plot.py:4462:28: PLR0124 Name compared with itself, consider replacing `max_ != max_`
  src/asammdf/gui/widgets/plot.py:4503:28: PLR0124 Name compared with itself, consider replacing `min_ != min_`
  src/asammdf/gui/widgets/plot.py:4505:28: PLR0124 Name compared with itself, consider replacing `max_ != max_`
  src/asammdf/gui/widgets/tabular_base.py:1382:16: PLR0124 Name compared with itself, consider replacing `target != target`
  Found 7 errors.
```

https://github.com/danielhrisca/asammdf/blob/43852ffec3707e835a5bf401851e65561620c944/src/asammdf/gui/widgets/plot.py#L136
https://github.com/danielhrisca/asammdf/blob/43852ffec3707e835a5bf401851e65561620c944/src/asammdf/gui/widgets/plot.py#L144
https://github.com/danielhrisca/asammdf/blob/43852ffec3707e835a5bf401851e65561620c944/src/asammdf/gui/widgets/plot.py#L4460
https://github.com/danielhrisca/asammdf/blob/43852ffec3707e835a5bf401851e65561620c944/src/asammdf/gui/widgets/plot.py#L4462
https://github.com/danielhrisca/asammdf/blob/43852ffec3707e835a5bf401851e65561620c944/src/asammdf/gui/widgets/plot.py#L4503
https://github.com/danielhrisca/asammdf/blob/43852ffec3707e835a5bf401851e65561620c944/src/asammdf/gui/widgets/plot.py#L4505
https://github.com/danielhrisca/asammdf/blob/43852ffec3707e835a5bf401851e65561620c944/src/asammdf/gui/widgets/tabular_base.py#L1382

### DIFF
--- a/src/asammdf/blocks/mdf_v4.py
+++ b/src/asammdf/blocks/mdf_v4.py
@@ -5693,15 +5693,13 @@ class MDF4(MDF_Common):
                             if size % 2:
                                 size += 1
                                 elem = elem + b"\0"
-                            data.append(UINT32_p(size))
-                            data.append(elem)
+                            data.extend((UINT32_p(size), elem))
                             off += size + 4
                     else:
                         for elem in signal:
                             offsets.append(off)
                             size = len(elem)
-                            data.append(UINT32_p(size))
-                            data.append(elem)
+                            data.extend((UINT32_p(size), elem))
                             off += size + 4
 
                     offsets = array(offsets, dtype=uint64)
@@ -7620,22 +7618,16 @@ class MDF4(MDF_Common):
                 )
                 vals = vals.view(types)
 
-                arrays = []
-                arrays.append(vals["ms"])
-                # bit 6 and 7 of minutes are reserved
-                arrays.append(vals["min"] & 0x3F)
-                # only firt 4 bits of hour are used
-                arrays.append(vals["hour"] & 0xF)
-                # the first 4 bits are the day number
-                arrays.append(vals["day"] & 0xF)
-                # bit 6 and 7 of month are reserved
-                arrays.append(vals["month"] & 0x3F)
-                # bit 7 of year is reserved
-                arrays.append(vals["year"] & 0x7F)
-                # add summer or standard time information for hour
-                arrays.append((vals["hour"] & 0x80) >> 7)
-                # add day of week information
-                arrays.append((vals["day"] & 0xF0) >> 4)
+                arrays = [
+                    vals["ms"],
+                    vals["min"] & 0x3F,  # bit 6 and 7 of minutes are reserved
+                    vals["hour"] & 0xF,  # only first 4 bits of hour are used
+                    vals["day"] & 0xF,  # the first 4 bits are the day number
+                    vals["month"] & 0x3F,  # bit 6 and 7 of month are reserved
+                    vals["year"] & 0x7F,  # bit 7 of year is reserved
+                    (vals["hour"] & 0x80) >> 7,  # add summer or standard time information for hour
+                    (vals["day"] & 0xF0) >> 4,  # add day of week information
+                ]
 
                 names = [
                     "ms",
@@ -10329,7 +10321,7 @@ class MDF4(MDF_Common):
 
         dbc = None
 
-        for i, channel in enumerate(channels):
+        for channel in channels:
             if channel.name == "CAN_DataFrame":
                 attachment_addr = channel.attachment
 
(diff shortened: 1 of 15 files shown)


danielhrisca:
Those comparison are to check is the value is not a number NaN

zariiii9003:
> Those comparison are to check is the value is not a number NaN

Ah, i didn't know that. Would you mind if i replace it with np.isnan for readability?

danielhrisca:
My concern is the execution speed in this case

image

src/asammdf/blocks/mdf_v4.py
```diff
@@ -5693,15 +5693,13 @@ def extend(self, index: int, signals: list[tuple[NDArray[Any], NDArray[Any] | No
...
-                            data.append(elem)
+                            data.extend((UINT32_p(size), elem))
                             off += size + 4
                     else:
                         for elem in signal:
                             offsets.append(off)
                             size = len(elem)
-                            data.append(UINT32_p(size))
-                            data.append(elem)
+                            data.extend((UINT32_p(size), elem))
```
danielhrisca, line 5702: Is this automatically done by ruff?

src/asammdf/blocks/mdf_v4.py
```diff
@@ -6593,7 +6590,7 @@ def get(
             if name is None:
                 name = channel.name
 
-            unit = conversion and conversion.unit or channel.unit
+            unit = conversion.unit if conversion else channel.unit
```
danielhrisca, line 6593: This is not right: if the conversion unit is empty then we need to check the channel unit as well

src/asammdf/gui/dialogs/advanced_search.py
```diff
@@ -174,7 +174,7 @@ def search_text_changed(self):
                                     (group_index, channel_index): {
                                         "names": [ch.name],
                                         "comment": extract_xml_comment(ch.comment).strip(),
-                                        "unit": ch.conversion and ch.conversion.unit or ch.unit,
+                                        "unit": ch.conversion.unit if ch.conversion else ch.unit,
```
danielhrisca, line 177: same as above, also applies to the following commit diffs

src/asammdf/gui/widgets/tabular_base.py
```diff
@@ -1379,30 +1379,20 @@ def apply_filters(self, event=None):
             column_name = pandas_query_compatible(column_name)
             op = filter.op.currentText()
 
-            if target != target:
-                # here we have NaN
-                nan = np.nan
```
danielhrisca, line 1384: The pandas query check the local scope, so this has to remain otherwise `@nan` will fail

danielhrisca (changes requested):
please address the open topics

src/asammdf/blocks/mdf_v4.py
zariiii9003: Not really, this actually not part of the pylint rules, instead i ran some of the preview rules: https://docs.astral.sh/ruff/rules/repeated-append/
Even then this is only done automatically if you call ruff with `--unsafe-fixes`. 
Should i revert these?

danielhrisca: It's ok I guess, it was more a curiosity

src/asammdf/blocks/mdf_v4.py
zariiii9003: True. I just found out, that the preview rule, which flagged this, will be [removed soon](https://github.com/astral-sh/ruff/issues/9007)

zariiii9003: Fixed

src/asammdf/gui/dialogs/advanced_search.py
zariiii9003: Fixed

src/asammdf/gui/widgets/tabular_base.py
zariiii9003: Fixed

danielhrisca (approved):

PR #978, opened 2024-01-30
Parsing of nested CA structures

VorobiovM:
Extend `def _read_channels` to read nested CN blocks from within CA composition.

### DIFF
--- a/src/asammdf/blocks/mdf_v4.py
+++ b/src/asammdf/blocks/mdf_v4.py
@@ -17,6 +17,7 @@ from math import ceil, floor
 import mmap
 import os
 from pathlib import Path
+import re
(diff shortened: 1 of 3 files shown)


danielhrisca:
This is a step forward compared to the current code. Do you think you can get to have actual field values for the referenced structs? right now the code just returns the bytes

VorobiovM:
Sorry for not communicating quicker, I've found a bug with sources and a bug with entries for nested channels.
Still I am not sure about what you mean, when selecting signals I have correct data type returning

mesa-git:
Hey guys,

any chance we can speed up this PR a bit? :) the changes are really important.

PR #996, opened 2024-03-15
Isotp

hacklint:
I added basic ISO-TP support ("fair weather" - assumes partial frames arrive in order etc).
Development done based on a small number of example files and it does solve my problem.
Providing it here as-is, glad if it is useful for others.
Basic test of actual frame merging, but not on signal level (I look for presence of associated flow control
frames in the DBC to determine if we have ISO-TP or not. Not sure if this is the best way, but it
works for me (TM) :)


danielhrisca:
thanks @hacklint

PR #997, opened 2024-03-15
Warn and relax global j1939 if non-extended IDs present

hacklint:
I had a DBC file containing one message with non-extended ID.
asammdf refused to accept that and threw an exception even if the rest of the file was OK.
I think warning in this case may be a less drastic approach - the suggested change allows me
to run without modifying either asammdf locally _or_ my incoming data, while still passing the tests.

Speaking of tests - it seems to me that running the GUI tests on a large screen (3840 x 2160)
places the gui window in the lower right part of the screen, but it seems the cursor is placed farther 
and farther to "north-west", ie up and left - and after a while leaving the window which stops the
test sequence. Manually moving it back makes tests continue, but the behaviour is repeated,
Running on a "standard" sized screen does not seem to behave like this.

### DIFF
--- a/src/asammdf/mdf.py
+++ b/src/asammdf/mdf.py
@@ -4755,6 +4755,14 @@ class MDF:
             messages = {(message.arbitration_id.id, message.arbitration_id.extended): message for message in dbc}
 
             global_is_j1939 = dbc.attributes.get("ProtocolType", "").lower() == "j1939"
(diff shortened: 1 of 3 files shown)


hacklint:
Realised J1939 shall allow 11-bit ID traffic on same CAN bus so I created https://github.com/danielhrisca/asammdf/issues/1008

PR #1035, opened 2024-06-10
Use temporary folder in MDF4

alex-ruehe:
Use the `temporary_folder` value when copying the MDF4 file to a temporary location and use `gettempdir()` as fallback.


danielhrisca:
thank you

PR #1065, opened 2024-08-09
Fix reading of CA-CN structures

VorobiovM:
Inside a CA-CN structure, if last element is a CA block no new channels will be created out of CA.
Also, modified `_read_channel` to not replace original channel name in-place, instead a new channel is created.

P.S. Updated python version in CONTRIBUTING.md

### DIFF
--- a/src/asammdf/blocks/mdf_v4.py
+++ b/src/asammdf/blocks/mdf_v4.py
@@ -1044,6 +1044,13 @@ class MDF4(MDF_Common):
                     first_dep = ca_block = ChannelArrayBlock(address=component_addr, stream=stream, mapped=mapped)
                     dependencies[index] = [first_dep]
 
+                    ca_cnt = len(dependencies[index])
+                    byte_offset_factors = []
+                    bit_pos_inval_factors = []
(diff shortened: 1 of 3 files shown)


VorobiovM:
@danielhrisca do you know why __isal__ is failing to build?

VorobiovM:
@danielhrisca I think only UI test fail, could you take a look?

src/asammdf/blocks/mdf_v4.py
```diff
@@ -6576,8 +6568,6 @@ def attach(
 
         file_path = Path(file_name)
 
-        mime: str | None
-
         mime, _ = guess_type(Path(file_path))
```
JulienGrv, line 6571: Hi @VorobiovM, I had this error as well with mypy. I fixed it differently though. It's unclear to me why mime is a function argument although it gets overridden line 6571.

@danielhrisca did you keep the function argument mime for backward compatibility and will remove it in some future version?

SGHAStefan:
Hi @danielhrisca , I have fixed some issues on the branch. Can you please consider this branch for merging?

PR #1066, opened 2024-08-10
Add import for retrieving zips

artyeffem:
Retrieving zips with urlllib requires explicit import of its request module


danielhrisca:
thank you

PR #1071, opened 2024-08-23
Changes to the loading of vlsd values

JanvonScholley:
Was not able to read single vlsd value which was split between two signal data blocks. Or multiple vlsd values when the last vlsd value was split between two signal data blocks. This resulted in asammdf ERROR "samples and timestamps length mismatch( x vs x+1). With this changes those values should be able to be loaded without causing an error.

### DIFF
--- a/src/asammdf/blocks/mdf_v4.py
+++ b/src/asammdf/blocks/mdf_v4.py
@@ -1307,10 +1307,55 @@ class MDF4(MDF_Common):
                         elif block_type == v4c.DZ_BLOCK_LZ:
                             new_data = lz_decompress(new_data)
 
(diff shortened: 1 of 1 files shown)


danielhrisca:
Hello @JanvonScholley ,

thanks for the PR. Looking over your changes I felt that it could be done in a simpler way so I made the code changes found tin this development branch commit https://github.com/danielhrisca/asammdf/commit/117550ecda94aaff3fd674c09c22cc3eb1707842

Please let me know if it works correctly with you files

JanvonScholley:
Hello @danielhrisca,

thank you for the fix. This also works for me.

PR #1093, opened 2024-11-07
enable numpy 2

Simsalabim1234:


danielhrisca:
My biggest concern with numpy 2.0 is the new Python integer handling that can cause problems

```python
>>> import numpy as np
>>> np.arange(4, dtype='u1') * (-5)
  File "<stdin>", line 1, in <module>
OverflowError: Python integer -5 out of bounds for uint8
>>> np.arange(4, dtype='u1') * 1247
  File "<stdin>", line 1, in <module>
OverflowError: Python integer 1247 out of bounds for uint8
```

Simsalabim1234:
> My biggest concern with numpy 2.0 is the new Python integer handling that can cause problems
> 
> ```python
> >>> import numpy as np
> >>> np.arange(4, dtype='u1') * (-5)
> Traceback (most recent call last):
>   File "<stdin>", line 1, in <module>
> OverflowError: Python integer -5 out of bounds for uint8
> >>> np.arange(4, dtype='u1') * 1247
> Traceback (most recent call last):
>   File "<stdin>", line 1, in <module>
> OverflowError: Python integer 1247 out of bounds for uint8
> ```

Hmm.. I don't have an overview over the codebase, but to me this does not look like a hard blocker, but more like an opportunity to introduce some checks and explicit casts to make the code more stable.

danielhrisca (approved):

PR #1097, opened 2024-11-16
Feature/cleanup and fix issues

JulienGrv:
What has been done:

- update GitHub actions
- use editable install, requirements are placed in the setup.py and it is not required to repeat them anywhere else
- use `python -m build` in place of the deprecated `python setup.py`
- shell snippets are POSIX compliant now
- update pre-commit hooks
- remove `version: 3.8` in .readthedocs.yml because it does not work
- add VSCode recommended extensions to work on the project
- add developer instructions in CONTRIBUTING.md
- update README.md
- format markdown, yaml and toml files
- remove deprecated setup.cfg, pylint.cfg and pre_commit.bat
- align ruff and black settings between direct usage and pre-commit usage
- always install decode and export extra dependencies for gui extra
- fix issue where python builtins were deleted and overriden, which caused python internals to fail when looking for the `eval` function. Unfortunately, fixing this issue introduced another one for the gui. However, as a general rule of thumb, we should not mess with python builtins or else we can break python itself. Hopefully we can find a better implementation in the future.
- remove `fastpath` argument to pandas Series as it is long deprecated and causes warnings when running purest
- tox.ini settings could be simplified as well thanks to new canonical requirements.txt files and gui extras installation


src/asammdf/mdf.py
```diff
@@ -176,26 +176,26 @@ class MDF:
...
+    use_display_names (**kwargs) : bool
         keyword only argument: for MDF4 files parse the XML channel comment to
         search for the display name; XML parsing is quite expensive so setting
         this to *False* can decrease the loading times very much; default
         *True*
-    remove_source_from_channel_names (\*\*kwargs) : bool
-        remove source from channel names ("Speed\XCP3" -> "Speed")
-    copy_on_get (\*\*kwargs) : bool
+    remove_source_from_channel_names (**kwargs) : bool
+        remove source from channel names ("SpeedXCP3" -> "Speed")
```
danielhrisca, line 185: Here we have to keep the slash because i's the way some applications were embedding the source name in the channel name

danielhrisca:
Let me know when it is ready

danielhrisca (approved):

danielhrisca:
thank you Julien!

PR #1107, opened 2024-12-05
fix tz_offset being unsigned int

hutauf:
fix tz_offset being unsigned int, however it's supposed to be a signed int

In the MDF spec for v3:
INT16 1 UTC time offset in hours (= GMT time zone)

The time offset is defined as INT16, not UINT16.

If the time offset is negative, but interpreted as UINT16, it will result in a very large positive number.


danielhrisca:
Thank you @hutauf !

PR #1113, opened 2024-12-06
Move project configuration from setup.py to pyproject.toml / Build ABI3 wheels

zariiii9003:


zariiii9003:
Hi Daniel, you can use the [build package](https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#packaging-your-project)

```shell
pip install build
python -m build
```

danielhrisca:
thanks

danielhrisca:
the pyinstaller build is not packaging `cutils.pyd` since this PR. Any idea how to fix it?

zariiii9003:
I'll take a look.

JulienGrv:
Just leaving this here for reference: https://scikit-build-core.readthedocs.io/en/stable/configuration.html#editable-installs

MatinF:
The merge of this PR into the development branch results in the GUI build to not work with the error below:

```
  File "asammdfgui.py", line 52, in <module>
    main()
  File "asammdfgui.py", line 29, in main
    from asammdf.gui.utils import excepthook, set_app_user_model_id
  File "PyInstaller\loader\pyimod02_importers.py", line 384, in exec_module
  File "asammdf\__init__.py", line 14, in <module>
  File "PyInstaller\loader\pyimod02_importers.py", line 384, in exec_module
  File "asammdf\blocks\source_utils.py", line 12, in <module>
  File "PyInstaller\loader\pyimod02_importers.py", line 384, in exec_module
  File "asammdf\blocks\v4_blocks.py", line 45, in <module>
ModuleNotFoundError: No module named 'asammdf.blocks.cutils'

```
Is there a way to avoid this by adding a change to this PR? If not, can this PR be reverted so that it is not merged into the development branch? Currently this is blocking our ability to download and host the latest asammdf development branch GUI. It would be really great to have this resolved.

I believe this also relates to Daniel's comment on Dec 10:

```
the pyinstaller build is not packaging cutils.pyd since this PR.
```

zariiii9003:
@MatinF Did you try to build via `tox -e pyinstaller`?

MatinF:
@zariiii9003 My motivation is to download a version of the latest development GUI, which we host on our website for our end users. Therefore it needs to be 'plug & play' without any additional build/installation involved. 

Usually we achieve this by getting the latest dev build from below:
https://github.com/danielhrisca/asammdf/actions/workflows/build_windows.yml

However, this is not currently up-to-date with the latest changes.

The pyinstaller builds you link to currently have the issue on my PC (and therefore probably many of our users) that it comes up as a virus, causing Windows to delete the exe file. This is on Windows 10:
image

I don't see a way to get around this, but to me it seems that the pyinstaller is not a viable way for us to share the GUI with the thousands of end users we work with that will need to be able to open it without issues, hence I was hoping for the other way to work - which it does, except for after the changes that were made with this specific PR, which breaks the GUI with the warning message I listed.

MatinF:
Hi again,

When I open the link I shared, it looks as below where I select the `build_windows`:
image

I then go to the latest build as per the arrow and download the dist build:
image

This normally works - and it works without the false positive I get when trying the alternative pyinstaller route. However, since this cmake PR was incorporated into the development branch, this method fails as the GUI throws the error I listed. 

The cxfreeze link you sent does also seem to work - so if this is something that could be incorporated as a way to resolve the impact of the cmake change that would be ideal. But I guess part of my challenge is that only the pyinstaller version seems to be continuously updated, with the last update to the `build_windows` GUI builds appears to be 2 months old. So I am a bit unsure how we can get the latest development builds via this method, even if the cmake issue is resolved?

zariiii9003:
For clarification: the "build_linux" and "build_windows" workflows were unified into "pyinstaller_build" since they were doing the same things anyway. The artifact names were changed from "dist" to e.g. "pyinstaller-windows-latest" to reflect the build tool and target operating system.
But the old workflows were using pyinstaller, too. Nothing changed there.

MatinF:
Ok thanks a lot for clarifying, this makes it much more clear to me. It looks like  when I try the latest build via the pyinstaller, the GUI does open (although there is an error, but not related to the virus aspect). I will test this out further.

PR #1115, opened 2024-12-10
Get bits

danielhrisca:


zariiii9003:
I see you're getting a linker error with Python 3.9, because `Py_NewRef` was added to stable ABI in 3.10.

You can change the configuration, so python 3.9 builds a version specific wheel and 3.10+ builds an ABI3 wheel.
Change this line in pyproject.toml: `wheel.py-api = "cp310"` and change your CMakeLists.txt like this:

```CMake
if(NOT "${SKBUILD_SABI_COMPONENT}" STREQUAL "")
    set(PY_ABI_OPTIONS "WITH_SOABI" "USE_SABI" "3.10")
else()
    set(PY_ABI_OPTIONS "")
endif()
python_add_library(cutils MODULE src/asammdf/blocks/cutils.c ${PY_ABI_OPTIONS})
```

PR #1116, opened 2024-12-10
Fix Pyinstaller Build

zariiii9003:
I think the problem is the "editable" install with scikit-build-core. Building the package without build isolation would be necessary, so the pyd file is placed into the src directory.

So this is more of a workaround than a fix: Within the toxenv asammdf is installed from sdist, so the behavior of editable installs is irrelevant.

I also merged the three build workflows into one. Unless there was a reason to keep them separate...


danielhrisca:
thanks a lot!

danielhrisca:
Ping @MatinF

JulienGrv:
Just leaving this here for reference: https://scikit-build-core.readthedocs.io/en/stable/configuration.html#editable-installs

PR #1118, opened 2024-12-11
Fix readthedocs "Config file validation error"

zariiii9003:
The last successful documentation build is for version 7.3.*

Lately the build fails with
image


danielhrisca:
You are on a roll! thank you!

PR #1136, opened 2025-01-26
Hotfix/checkout fails in ci

JulienGrv:
Git checkout is failing in CI for branches in fork repositories but also for PR somehow because the token `SUBMODULE_PAT` only exists for the asammdf repository. I removed the token because it should not be needed as libdeflate is a public repository (tested in my repository and it worked) or is there any reason why it might be needed that I overlooked?

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -13,8 +13,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
         with:
-           submodules: 'recursive'
-           token: ${{ secrets.SUBMODULE_PAT }}
+          submodules: "recursive"
 
       - name: Set up Python
         uses: actions/setup-python@v5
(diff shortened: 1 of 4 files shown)


danielhrisca:
From what I saw the submodule was not properly retrieved in the github actions jobs without the token

JulienGrv:
Test pipelines are passing so it works without token. I have to check for the other pipelines. If you look at my other PR where I did not edit the main.yml, you will see that test pipelines are failing because the token is required, which does not work because I don't have a SUBMODULE_PAT token defined in my repository settings. I could add it but it's not necessary because libdeflate is a public repository.

I will look at the other pipelines though. Maybe we need to update to checkout v4.

I will come back to you with more info

danielhrisca:
let me know when your pull requests are ready

JulienGrv:
@danielhrisca this one is ready to merge thank you

danielhrisca:
thanks

PR #1147, opened 2025-02-09
Corrected misleading typo in ChannelConversion block's comment

Tyronnosaurus:
The code for the ChannelConversion block works well but one of its comments has a small typo contradicting the standard.
The docs are autogenerated from the code so they also show this typo.


danielhrisca:
Thanks, it has been there for ages

PR #1150, opened 2025-02-16
Format with ruff instead of black

morbult:
This PR replaces black with ruff's formatter. Ruff is already used for linting so why not use one tool instead of two here.

Line length is kept at 120.

As seen in ee60e89, some things are formatted a little differently although ruff is designed as a drop-in replacement for black. 
Some deviations are known and I think in this case these are all the deviations happening:
* https://docs.astral.sh/ruff/formatter/black/#blank-lines-at-the-start-of-a-block
* https://docs.astral.sh/ruff/formatter/black/#f-strings
* https://docs.astral.sh/ruff/formatter/black/#implicit-concatenated-strings
* https://docs.astral.sh/ruff/formatter/black/#call-chain-calls-break-differently-in-some-cases

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,8 +1,4 @@
 repos:
-  - repo: https://github.com/psf/black-pre-commit-mirror
-    rev: 25.1.0
-    hooks:
-      - id: black
   - repo: https://github.com/astral-sh/ruff-pre-commit
     rev: v0.9.5
     hooks:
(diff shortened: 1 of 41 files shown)


danielhrisca:
@morbult thanks, let me know when it is ready

morbult:
It is ready. Sorry that all the commits from #1149 were included, it was only meant to be the last three. Still learning the Github workflow :)

morbult:
@danielhrisca Should I create a new PR with just the relevant commits?

.pre-commit-config.yaml
JulienGrv, line 1: Here you can add the format command for the ruff hook I think

morbult:
See #1151

PR #830, opened 2023-03-09
Update signal.py

kevinsmia1939:
This will fix this issue.
https://github.com/danielhrisca/asammdf/issues/829

Similar issue report here.
https://github.com/Udayraj123/OMRChecker/pull/56

PR #1181, opened 2025-03-27
Fix 'FigureCanvasTkAgg' object has no attribute 'set_window_title'

morbult:
This was partially fixed in #830 but two places were missed.


danielhrisca:
Thanks

PR #1184, opened 2025-03-29
Fix import errors in integration tests

JulienGrv:
I am not qualified to fix the GUI tests but I could at least fix the import errors


danielhrisca:
Thanks Julien

PR #1186, opened 2025-04-03
fix(mdf_v4): wrong size when decompressing DZ block

JulienGrv:
compressed_size is used instead of original_size when calling decompress function


JulienGrv:
@danielhrisca test seems to still be failing so I am investigating further. I will come back to you

danielhrisca (approved):

PR #1192, opened 2025-04-10
Fix tests

FillBk:
In this PR:
- [x] was updated tests
- [x] fixed batch issue when user cannot select file for batch modify and export feature
- [x] fix trim_info width issue


FillBk:
Zooming action result a strange bug. This is `trim_info` of signal. 

```
E       First differing element 2:
E       200
E       199.5
E       
E       - (np.float64(2.642860623448), np.float64(9.722452976555), 200)
E       ?                                                          ^^^
E       
E       + (np.float64(2.642860623448), np.float64(9.722452976555), 199.5)
E       ?    
```

PR #1151, opened 2025-02-17
Format with ruff instead of black

morbult:
This PR replaces black with ruff's formatter. Ruff is already used for linting so why not use one tool instead of two here.

Line length is kept at 120.

As seen in ee60e89, some things are formatted a little differently although ruff is designed as a drop-in replacement for black.
Some deviations are known and I think in this case these are all the deviations happening:

* https://docs.astral.sh/ruff/formatter/black/#blank-lines-at-the-start-of-a-block
* https://docs.astral.sh/ruff/formatter/black/#f-strings
* https://docs.astral.sh/ruff/formatter/black/#implicit-concatenated-strings
* https://docs.astral.sh/ruff/formatter/black/#call-chain-calls-break-differently-in-some-cases

PR #1207, opened 2025-05-27
Format with ruff instead of black

morbult:
Same as #1151 but no conflicts

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,7 +1,7 @@
 repos:
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.11.10
+    rev: v0.11.11
     hooks:
-      - id: ruff
+      - id: ruff-check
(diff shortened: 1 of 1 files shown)


.pre-commit-config.yaml
```diff
@@ -1,10 +1,7 @@
 repos:
-  - repo: https://github.com/psf/black-pre-commit-mirror
-    rev: 25.1.0
-    hooks:
-      - id: black
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.11.8
+    rev: v0.11.10
```
zariiii9003, line 3: v0.11.11 is the latest. You can also replace `id: ruff` with `id: ruff-check`

morbult: Good point. I used the hook from https://docs.astral.sh/ruff/integrations/#pre-commit, but now it's fixed.

PR #1211, opened 2025-06-01
Bugfix/mypy-errors

JulienGrv:
There was some new type errors that I fixed. We are very close to having a running CI again, there is only one test that is still failing on Ubuntu. Hopefully we can get it fixed soon. 

Notice, there was a series of functions and classes in `blocks.utils` that are using PySide6 types and only used for the GUI so I moved them into the gui module as I don't want to maintain type annotations for the GUI part. I put them under the package `gui.serde` as it seems like a good name since it is related to serialization and deserialization as I understood from reading the code. If you don't like the name `serde`, just let me know and I will rename or rearrange it to whatever you prefer


src/asammdf/blocks/v4_blocks.py
```diff
@@ -4719,7 +4722,7 @@ def __init__(self, **kwargs: Unpack[DataZippedBlockKwargs]) -> None:
 
     def __setattr__(self, item: str, value: object) -> None:
         if item == "data" and not self._prevent_data_setitem:
-            data = typing.cast(bytes, value)
+            data: bytes | memoryview[int] = typing.cast(bytes, value)
```
JulienGrv, line 4725: - [ ] check value type for item == "data"

src/asammdf/blocks/mdf_v4.py
```diff
@@ -1361,7 +1361,7 @@ def _load_signal_data(
                 else:
                     start_offset, end_offset = offsets
 
-                    data = bytearray()
+                    data_array = bytearray()
```
JulienGrv, line 1364: Make sure data is not unbound when returning

danielhrisca:
is this ready for merging?

JulienGrv:
Yes it's ready

danielhrisca:
thanks Julien!

PR #1218, opened 2025-06-16
Investigate `CI` tests failures

FillBk:
In this PR:
- used HD display (1280x720);
- move gui test `screenshots` folder to the same address as `resources` and `test_workspace`;
- save widget screenshots on `tearDown` from all platforms and python versions
- investigate `self.widget.grab` only if failure, but `pytest._outcome` wraps `unittest._outcome` and for now is not reliable to modify all tests and add `pytest.hookimpl`;
- investigate and repair failed tests.

### DIFF
--- a/test/asammdf/gui/test_base.py
+++ b/test/asammdf/gui/test_base.py
@@ -63,7 +63,7 @@ class TestBase(unittest.TestCase):
 
     resource = os.path.normpath(os.path.join(os.path.dirname(__file__), "resources"))
     test_workspace = os.path.join(os.path.dirname(__file__), "test_workspace")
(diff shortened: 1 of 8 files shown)


test/asammdf/gui/test_base.py
```diff
@@ -114,7 +114,11 @@ def setUp(self) -> None:
         if not os.path.exists(self.screenshots):
             os.makedirs(self.screenshots)
 
-        os.makedirs(self.test_workspace)
+        try:
+            os.makedirs(self.test_workspace)
```
danielhrisca, line 118: https://docs.python.org/3/library/os.html#os.makedirs `exist_ok=True`

danielhrisca (approved):

PR #1223, opened 2025-06-23
enable using distro's libdeflate

a-detiste:
Hi,

I'm packagig asammdf for Debian. This patch would make long time maintenance much easier.

The current behaviour is kept by default.


danielhrisca (approved):

a-detiste:
Indeed it's a long term commitment.

PR #1243, opened 2025-09-15
Respect `use_interpolation` with `use_polars`

kapinga:
The existing design of `MDF.to_dataframe` interpolates signal readings whenever `use_polars` is True, because Polars `Series` objects do not have an index and are not automatically aligned in the final DataFrame creation. For my purposes (usually a forensic examination of CAN bus data), it is helpful to disable interpolation, but I want to leverage polars while doing so.

This PR includes changes to the operation of `MDF.to_dataframe`. In particular, the local variable `data` collects two-column DataFrames (rather than Series objects) when `use_polars` is set to True. This ensures that we collect the same data (timestamp and signal) as in the code path when `use_polars` is False.

At the end of the function, these individual frames are then (lazily) joined on the `timestamps` column, resulting in the final DataFrame.

We get a couple of side benefits:

* Checks are also removed for `use_polars` in combination with `interpolation_outwards_with_nan`. This check is no longer needed as we now support null-filling with polars.
* As the code around handling `time_as_date` already was changed, I cleaned things up to avoid calling `pd.to_timedelta` and removed the need for the `# type: ignore[assignment]` comments. After testing, the new code produces timestamps within 1 ns of the results using `pd.to_timedelta` on real data files.

### DIFF
--- a/src/asammdf/mdf.py
+++ b/src/asammdf/mdf.py
@@ -5359,9 +5359,12 @@ class MDF:
             mdf.close()
             return result
 
+        if use_polars and not POLARS_AVAILABLE:
+            raise MdfException("to_dataframe(use_polars=True) requires polars")
+
         target_byte_order = "<=" if sys.byteorder == "little" else ">="
 
-        data: dict[str, NDArray[Any] | pd.Series[Any]] | dict[str, pl.Series] = {}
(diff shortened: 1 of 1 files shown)


danielhrisca (approved):
thank you for your contribution!

PR #1266, opened 2025-12-01
Fix Qt warnings

neomilium:
If we let the `self`, `asammdf` warns on console about:

```
QLayout: Attempting to add QLayout '' to BusDatabaseManagerDialog BusDatabaseManagerDialog, which already has a layout
```


danielhrisca (approved):

danielhrisca:
thanks @neomilium

PR #1267, opened 2025-12-01
Fix inconsistent CAN unknown IDs tracking

neomilium:
In `src/asammdf/blocks/mdf_v4.py`, the `_extract_can_logging` method uses inconsistent keys when tracking unknown CAN message IDs:

- **Line 10134**: When NOT found: `unknown_ids[msg_id].append(True) `- uses `msg_id` (int)
- **Line 10158**: When found: `unknown_ids[(msg_id, is_extended)].append(False) `- uses `(msg_id, is_extended)` (tuple)

This causes a message to have separate dictionary entries for found vs not-found states, breaking the `all(not_found)` check on line 10316.

### DIFF
--- a/src/asammdf/blocks/mdf_v4.py
+++ b/src/asammdf/blocks/mdf_v4.py
@@ -10140,7 +10140,7 @@ class MDF4(MDF_Common[Group]):
                                         message = _msg
                                         break
                                 else:
-                                    unknown_ids[msg_id].append(True)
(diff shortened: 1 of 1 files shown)


danielhrisca:
please see the conflicts

neomilium:
I  moved the commit on top of `development` branch

danielhrisca:
thanks @neomilium

PR #1268, opened 2025-12-06
Fix configuration save with GPS window

neomilium:
The root cause was that `runJavaScript()` was being called with an incorrect number of parameters. The PySide6 `QWebEnginePage.runJavaScript()` signature is:
```python
page.runJavaScript(script, callback)
```
But the code was passing:
```python
page.runJavaScript(script, self.map.mapWidgetIndex,
```
This caused the callback to not be properly registered, leading to result getting an empty string or not being called correctly.
The fix:
 - Removed the `self.map.mapWidgetIndex` parameter
 - Added filtering in the callback to reject None or empty string values

Now it should properly wait for a valid zoom value or timeout and returns the default of 15.

### DIFF
--- a/src/asammdf/gui/widgets/gps.py
+++ b/src/asammdf/gui/widgets/gps.py
@@ -91,10 +91,11 @@ class GPS(Ui_GPSDisplay, QtWidgets.QWidget):
         result = []
 
         def callback(*args):
-            result.append(args[0])
+            if args and args[0] is not None and args[0] != '':
+                result.append(args[0])
 
         map_widget = self.map.getMapWidgetAtIndex(self.map.mapWidgetIndex)
-        map_widget.page.runJavaScript("map.getZoom()", self.map.mapWidgetIndex, callback)
+        map_widget.page.runJavaScript("map.getZoom()", callback)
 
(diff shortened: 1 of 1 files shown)


danielhrisca:
there are 2 overloads for the `runJvaScript` function

https://doc.qt.io/qt-6/qwebenginepage.html#runJavaScript-1

neomilium:
OK, I missed it.

```
void QWebEnginePage::runJavaScript(const QString &scriptSource, quint32 worldId = 0, const std::function<void (const QVariant &)> &resultCallback = {})
```

Reading the description, we can keep this PR as the `worldId` should be a value of Enum:

> To avoid conflicts with other scripts executed on the page, the world in which the script is run is specified by worldId.
> The world ID values are the same as provided by [QWebEngineScript::ScriptWorldId](https://doc.qt.io/qt-6/qwebenginescript.html#ScriptWorldId-enum), and between 0 and 256. If you leave out the world ID, the script is run in the MainWorld.

while `self.map.mapWidgetIndex` does not seems to fit the description.

Please note that without this patch the configuration save failed if a GPS window is displayed.

danielhrisca (approved):

PR #1275, opened 2026-03-01
Add name/source filter to channel/filter tree (Natural Sort mode only)

Bubba74:
Still work-in-progress, but most of the content is up and I think ready for discussion.

- Adds new user input boxes for specifying a smart-case regex search pattern for signal names / sources in the "Channels" and "Modify & Export" views.
  - I added the filters to "Modify & Export" for consistency, but I personally don't care because I don't use that workflow.
- Limits tree size to 10k signals. Profiling showed that it would take ~1-2 seconds to _clear_ the widget of all signals (for very large traces of course). Subsequent updates are fast, since the widget only needs to clear the previously filtered signals.
- Adds the "path" / "source" to the signal name to assist in differentiating between signals with the same name. Perhaps this should be updated to be the group name?

WIP:
- add some tests
- move 10k signal limit to user configuration files?
- add key-bind (I like Alt-F) to move focus to the search bar
- come up with a better solution for displaying a signal's path

Example Images:
image
image

### DIFF
--- a/src/asammdf/gui/ui/file_widget.py
+++ b/src/asammdf/gui/ui/file_widget.py
@@ -69,6 +69,16 @@ class Ui_file_widget(object):
 
         self.channels_layout.addWidget(self.channel_view)
 
+        self.channels_search_name = QLineEdit(self.verticalLayoutWidget)
+        self.channels_search_name.setObjectName(u"channels_search_name")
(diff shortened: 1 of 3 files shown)


danielhrisca (approved):

danielhrisca:
thanks

PR #1301, opened 2026-06-25
fix: handle None channel comment when encoding metadata

jbanar-volvo:
﻿This ports the fix from #1296 to master.

Original change:
- src/asammdf/blocks/v4_blocks.py
- text = comment -> text = comment or ""

Reason:
PR #1296 was merged into development; this brings the same fix into master.


danielhrisca (approved):

danielhrisca:
@jbanar-volvo thanks!

Issue #1012, opened 2024-04-23
Signals not able to find using MDF object by loading the MF4 file


Panda-Engineer:
Hi Daniel,

I need your help to solve my issue when loading MF4 files using asammdf. 

I have an MF4 file that contains a large number of CAN-FD signals. I am able to read all the signals using CANoe, but when I try to use asammdf to read the file, I find that there are very few signals in the groups. Many signals not able to find in the groups or channels.
Could you please provide some guidance on what I should do next?
Thank you very much.

BR,

Pandas-engineer

danielhrisca:
Without the file I can only suspect that it contains bus logging data that is not supported (FlexRay, MOST, Eth ...)

Panda-Engineer:
Hi Daniel,

Thank you very much for your prompt response. As you know, it is difficult for me to share internal MF4 files. My MF4 file only contains CAN-FD signals and does not include data recorded from other bus systems like FlexRay, MOST, Ethernet, etc.

I have sent you an email. I've included excerpts from the groups information read from the current MF4 file and sent it to your email. Please check if it's useful. There are some groups' information that I'm indeed not sure if they contain signals that I cannot read.

Also, I have another idea. Should we also use the ARXML file to read the signal content?

For me, this is indeed a challenge.

Thanks again.

BR,

Pandas-engineer

PR #1305, opened 2026-07-09
AUTOSAR container-PDU extraction for CAN bus logging (supersedes #912)

### LINKED ISSUE #534: CAN-FD Frame: Signals from second PDU in a specific Frame are missing
# Python version
('python=3.9.5 (tags/v3.9.5:0a7dcbd, May  3 2021, 17:27:52) [MSC v.1928 64 bit '
 '(AMD64)]')
'os=Windows-10-10.0.19042-SP0'
'numpy=1.20.3'
'asammdf=6.2.0'
```
# Code 

  
# Description

The fastest way to debug is to have the original file. For data protection you can use the static 
method _scramble_ to scramble all text blocks, and send the scrambled file by e-mail.

_Please describe the issue here._


Hello,

I found a problem with CAN-FD.

I have a frame where there can be 2 PDUs in it at one given time.

When there are 2 PDUs in the Frame at once, I dont find the signals from the second PDU (and thus the second signal-group) in the file and therefore get an MdfException.

The signals are in the file, I checked with CANoe and vSignalizer.

The whole stuff is dynamic. There can be 2 PDUs in one frame and in the next frame, there is only one PDU of the 2 possible PDUs.

I THINK (I can´t proof) that the problem occures, when the very first frame in the file with that specific ID has two PDUs in it. 
In that case I get the MdfException.

In case, the very first frame only contains one PDU, it seems to be ok.
I don´t know, if it is possible to check if the signals are missing if there is a frame with 2 PDUs "later" in the file, after the first frame was ok.

Best regards

### LINKED ISSUE #854: CAN-FD Frame: Signals from PDUs are missing
# Python version
 python: 3.9
os- windows-10
asammdf: 7.3.10
  
# Description
_Please describe the issue here._

I have faced problem with CAN-FD.

Since our current network architecture follows CAN FD structure, I have a data which contains frame with more than one PDUs and these PDUs contains required important signals. 

But currently, when I applied **extract_bus_logging()** function, the signals which are presents in the PDUs are missing. It means, after applying extract_bus_logging(), I do not get PDU signals in output. 

Based on code tracking, PDU signal extraction is not added yet. Correct me, if I am wrong.

Best regards,

### LINKED ISSUE #883: CAN- FD support
# Description

_Please describe the issue here._

Hi @danielhrisca ,
I worked on the CAN-FD bus extraction in ASAMMDF and pushed my updates on this PR: https://github.com/kruti2107/asammdf/pull/1

Could you please review my changes and provide feedback when you have time? I'm eager to hear your thoughts on the implementation.

Thank you!

Best regards,
Kruti

kipp-ing:
## AUTOSAR container-PDU extraction for CAN bus logging

Clean-room reimplementation of dynamic (and static) AUTOSAR container-PDU
extraction, superseding the approach in #912.

### Review guide

- **What:** a real per-frame header walk for dynamic AUTOSAR container PDUs, plus
  static (header-less) containers, with one channel group per contained PDU.
  Fixes #534 and #854, closes #883.
- **Where to look:** all new logic lives in new functions in
  `blocks/bus_logging_utils.py` (`extract_pdus`, `_emit_pdu_signals`,
  `_contained_pdu_muxer`, `_signal_byte_extent`). `mdf_v4.py` only gains one
  routing branch: `is_pdu_container` goes to `extract_pdus`, everything else still
  goes to `extract_mux` unchanged. No changes to `utils.py`, `mdf.py` or
  `.coveragerc`.
- **Only shared-path change:** two small fixes in `extract_signal` that the
  `extract_mux` path also uses (details below). Happy to split them into their own
  PRs if that is easier to review.
- **Evidence:** 1 237 247 decoded values match `canmatrix.Frame.unpack` with 0
  mismatches. Against an independent recording of the same drive, 298 of 303
  signals agree on every sample. The 541 non-container signals are bit-identical
  to `master`. Details under "Validation on real measurements".
- **Tests:** fully offline, in `test/test_CAN_pdu_extraction.py`; the existing
  `test/test_CAN_bus_logging.py` still passes.
- **CI:** the red `tests` check comes from `development` itself. 116 GUI tests
  fail there with `'str' object has no attribute 'suffix'`, which #1315 fixes. On
  Linux the set of failing tests is identical to `development`'s, and all new
  tests pass.

### Why a rewrite of #912

#912's `extract_pdu` decoded contained-PDU signals directly from the whole
frame with **no per-PDU header walk**, so it only produced correct values when
a PDU sat at a fixed frame offset. Dynamic containers — a variable sequence of
contained PDUs whose byte offsets shift per frame — were mis-decoded. It also
returned a `list` with a non-standard entry tuple instead of the
`dict[entry, dict[name, ExtractedSignal]]` contract the channel-group machinery
consumes, and touched `utils.py` / `mdf.py` / `.coveragerc`.

### What this does

- **Dynamic containers**: real per-frame header walk — reads `Header_ID` /
  `Header_DLC`, advances a per-frame byte offset, vectorized across frames.
  Reference algorithm is `canmatrix.Frame.unpack`. Header geometry is derived
  from the header signals (short 24+8, long 32+32), not hardcoded. Both header
  fields are read **unsigned**: canmatrix's ARXML parser marks them signed, so a
  0xFF padding byte decodes as a DLC of −1 and walks the offset *backwards* over
  a padded tail, rescanning the frame misaligned and inventing contained PDUs
  out of padding.
- **PDU-payload-relative signals**: each contained PDU's byte-aligned payload
  slice is isolated, then the existing `extract_signal` applies unchanged.
- **Short transmissions**: a sender may transmit a contained PDU shorter than
  its declared size — the header DLC is the authority. Signals reaching past the
  transmitted length are flagged through `invalidation_bits` rather than
  reported as measured data.
- **Static (header-less) containers**: decoded straight from the full frame
  (canmatrix rebases their signal start bits to frame-relative via each PDU's
  `OFFSET`; `Frame.unpack` itself refuses static containers).
- **One channel group per contained PDU**, via the existing untouched
  channel-group machinery — PDU identity is carried in the entry's `muxer`
  slot; same return contract as `extract_mux`.
- **CAN-FD**: extraction is purely `CAN_DataFrame.DataBytes`-width driven, so
  the FD flag has no functional effect on decoding.
- **Two fixes in `extract_signal`** (shared with the `extract_mux` path):
  - a signed signal of standard bit width (8/16/32/64) at a non-byte-aligned
    offset was viewed as `i{std_size}` on the padded width instead of being
    sign-extended from its real bit width (e.g. an 8-bit signed field at bit
    offset 1 returned 225 instead of −31);
  - a signal too wide for any integer dtype is kept as a raw byte matrix, so
    two's complement cannot be applied to it. Real containers carry opaque
    216/288/400-bit blobs declared *signed*, and those raised `OverflowError`
    on the `1 << bit_count` mask.

### Footprint

Only `src/asammdf/blocks/bus_logging_utils.py` (new `extract_pdus`,
`_emit_pdu_signals`, `_contained_pdu_muxer`, `_signal_byte_extent`) plus routing
in `mdf_v4.py` (`is_pdu_container` -> `extract_pdus`, else `extract_mux`). No
changes to `utils.py`, `mdf.py`, or `.coveragerc`.

### Testing

Fully offline in `test/test_CAN_pdu_extraction.py`:

- dynamic containers vs the `canmatrix.Frame.unpack` oracle (BE/LE headers,
  0x00/0xFF padding, unique multi-PDU frames, bit-packed non-byte-aligned
  signed signals);
- static containers vs an equivalent flat frame carrying the same
  frame-relative signals;
- full `MDF.extract_bus_logging` end-to-end on a 32-byte container and on a
  genuine 64-byte CAN-FD container (EDL flag + DataLength set);
- a contained-PDU signal wider than 64 bits declared signed;
- a header DLC shorter than the declared PDU size.

The existing `test/test_CAN_bus_logging.py` (real OBD/J1939 data) still passes,
confirming no regression to the `extract_mux` path.

### Validation on real measurements

Validated against two real OEM CAN-FD bus logs (5.1 M and 2.3 M CAN frames) and
three production ARXML databases covering three CAN channels — 15 real container
messages in total.

**Against `canmatrix.Frame.unpack` as the oracle**, sample by sample:
**1 237 247 decoded values, 0 mismatches.**

**Message-based vs signal-based.** The same drive was recorded twice by the
logger: once as raw CAN frames, once as signals decoded on the fly by the logger
toolchain. Cross-checking the container decode against that second recording is
an oracle that shares no code with asammdf or canmatrix. Of 303 comparable
signals, **298 agree on every one of 114 251 samples**. The five that do not are
free-running sequence counters and CRCs whose sample instants differ between the
two recordings by more than the comparison window — same value range, same
cardinality.

**No regression to the existing path.** Running the full
`MDF.extract_bus_logging` on the first measurement, on `master` and on this
branch: all **541** non-container signals are bit-identical; on `master` the
container messages yield only `Header_ID`/`Header_DLC`, this branch adds **789**
real contained-PDU signals in 58 additional channel groups. A further **676 008**
non-container values were separately confirmed against the oracle.

### Related issues

- Fixes #534 — signals from the second PDU in a dynamic CAN-FD frame were missing (only the PDU at offset 0 decoded); the per-frame header walk resolves this.
- Fixes #854 — `extract_bus_logging()` returned no PDU signals ("PDU signal extraction is not added yet").
- Closes #883 — CAN-FD PDU extraction support (supersedes the approach in #912).
- Related to #1012 — CAN-FD signals missing from groups; likely the same cause, unconfirmed without the source file.

### Why this matters

asammdf is widely used for exactly this kind of analysis, and thank you for
maintaining it. AUTOSAR container PDUs are a common way for modern CAN-FD networks
to pack many signals into a single frame. Without extraction, such messages decode
to nothing but `Header_ID` / `Header_DLC` on `master`, so real measurements end up
with most of their signals missing. The request has come up repeatedly since 2021
(#534, #854, #883, #912). This PR aims to keep asammdf covering what people
actually record today. I'm happy to adjust anything to make it easier to review or
maintain.

### DIFF
--- a/doc/buslogging.rst
+++ b/doc/buslogging.rst
@@ -41,6 +41,39 @@ There multiple ways to address this channel in this situation:
 #. CAN bus ID, ASAM conformant message ID and short signal name, delimited by dot
 
     .. code:: python
-    
-        mdf.get('CAN1.CAN_DataFrame_123.EngineStatus')   
-        
+
+        mdf.get('CAN1.CAN_DataFrame_123.EngineStatus')
+
+
+Container I-PDUs
+================
+
+AUTOSAR **container I-PDUs** pack several smaller PDUs into a single CAN(-FD)
+frame. When the database describes a frame as a PDU container, each contained
+PDU is decoded and gets **its own channel group**, so its signals are addressed
+exactly like any other bus logging signal:
+
+.. code:: python
+
+    mdf.get('EngineStatus')                       # short signal name
+    mdf.get('CAN1.VehicleStatus.EngineStatus')    # container frame name
+
+Both container layouts are supported: *dynamic* containers, where every
+contained PDU is prefixed by a header (``Header_ID`` + ``Header_DLC``) and its
+offset therefore changes from frame to frame, and *static* containers, where the
+contained PDUs sit at fixed offsets.
+
+A sender may transmit a contained PDU shorter than the length declared in the
+database — the header DLC is the authority. Signals that reach past the
+transmitted length would otherwise be decoded out of the neighbouring PDU or out
+of the container padding, so those samples are marked invalid:
+
+.. code:: python
+
+    sig = mdf.get('SomeSignal')
+    sig.invalidation_bits    # None, or True where the bytes were not transmitted
+
+Two limitations are worth noting: a multiplexed *contained* PDU is not modelled
+by **canmatrix**, so it cannot be decoded, and container I-PDUs are only handled
+for CAN — not for LIN.
+
--- a/doc/pdu_container_extraction.md
+++ b/doc/pdu_container_extraction.md
@@ -4,6 +4,11 @@ Developer note for the container-PDU support in
 `src/asammdf/blocks/bus_logging_utils.py` (`extract_pdus`) wired into
 `MDF._extract_can_logging` (`mdf_v4.py`).
 
+The user-facing documentation is the "Container I-PDUs" section of
+`doc/buslogging.rst`. This note is not part of the built docs (sphinx is
+configured for `.rst` only) — it records *why* the decoder looks the way it
+does, for whoever touches it next.
+
 ## What
 
 AUTOSAR **container I-PDUs** pack several contained PDUs into one CAN(-FD)
--- a/doc/pdu_container_extraction.md
+++ b/doc/pdu_container_extraction.md
@@ -1,0 +1,101 @@
+# AUTOSAR PDU-container extraction (CAN bus logging)
+
+Developer note for the container-PDU support in
+`src/asammdf/blocks/bus_logging_utils.py` (`extract_pdus`) wired into
+`MDF._extract_can_logging` (`mdf_v4.py`).
+
+## What
+
+AUTOSAR **container I-PDUs** pack several contained PDUs into one CAN(-FD)
+frame. `extract_mux` only understood `is_multiplexed` frames, so container
+frames (`message.is_pdu_container`) were routed through it and mis-decoded.
+`extract_pdus` decodes them and gives **each contained PDU its own channel
+group**, reusing the existing channel-group machinery unchanged.
+
+Two container layouts are supported:
+
+- **Dynamic** — each contained PDU is prefixed by a per-PDU header
+  (`Header_ID` + `Header_DLC`). A PDU's byte offset depends on the lengths of
+  the PDUs before it, so the frame is walked header-by-header.
+- **Static** — no per-PDU header; contained PDUs sit at fixed byte offsets.
+
+## How
+
+`_extract_can_logging` routes `message.is_pdu_container` frames to
+`extract_pdus`; everything else keeps using `extract_mux`. `extract_pdus`
+branches on whether the frame exposes the `Header_ID`/`Header_DLC` synthetic
+signals that canmatrix injects.
+
+### Dynamic containers
+
+1. Header geometry is derived from the header signals (short header = 24 + 8
+   bits, long header = 32 + 32); it is **not** hardcoded.
+2. Headers are walked across all frames one "slot" per iteration. Frames
+   diverge in offset after the first PDU, so a per-frame byte offset is
+   advanced independently and each frame's current header window is gathered
+   vectorized. Unknown/padding ids advance by their DLC (matching canmatrix),
+   which guarantees termination.
+3. Per contained PDU, its payload slices are gathered into a 2-D array and its
+   signals extracted. A contained PDU's signal `start_bit` values are
+   **PDU-payload-relative**, so once the byte-aligned PDU payload slice is
+   isolated the regular `extract_signal` machinery applies unchanged.
+
+### Static containers
+
+canmatrix cannot decode static containers at runtime — `Frame.unpack` raises
+`DecodingConatainerPdu` on them. But its ARXML parser has already rebased each
+contained PDU's signal `start_bit` values to be **frame-relative** (via the
+PDU's `OFFSET`). So every contained PDU decodes straight from the full frame
+payload; each still becomes its own channel group. Static PDUs carry no header
+id (`pdu.id is None`), so the channel-group identity keys on the PDU name.
+
+### Shared emission
+
+Both paths call `_emit_pdu_signals(...)`, which extracts a single PDU's signals
+into a dedicated channel-group entry (the PDU identity is carried in the entry's
+`muxer` slot via `_contained_pdu_muxer`). The only difference is the payload it
+is handed: the sliced PDU payload (dynamic) or the full frame (static).
+
+### CAN-FD note
+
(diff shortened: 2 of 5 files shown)


kipp-ing:
### Update: validated against real measurements, three defects fixed

I got access to more real-world data — two OEM CAN-FD bus logs (5.1 M and 2.3 M CAN frames) with three production ARXML databases, 15 real container messages — and ran the decoder against it before asking anyone to merge this. Notably, the same drive exists both **message-based** (raw CAN frames) and **signal-based** (decoded on the fly by the logger toolchain), which gives an oracle that shares no code with asammdf or canmatrix.

That was worth doing: the in-memory tests were green, but real containers are messier than synthetic ones and turned up three defects that only real data reaches.

1. **`OverflowError` aborted the whole extraction.** Real containers carry opaque 216/288/400-bit blobs declared *signed*. Those are kept as a raw byte matrix, so `as_non_byte_sized_signed_int` computed `1 << 216` on them. Signals too wide for an integer dtype now skip two's complement, exactly like their unsigned counterparts already did.

2. **Container padding was reported as measured data.** A sender may transmit a contained PDU shorter than its declared size — the header DLC is the authority, not the database. Signals reaching past the transmitted length were being decoded out of the neighbouring PDU or the padding. They are now flagged through `invalidation_bits`. Real data hits this on 2.4 % of contained-PDU occurrences.

3. **The header walk could run backwards.** canmatrix's ARXML parser marks the synthetic `Header_ID`/`Header_DLC` signals *signed*, so a 0xFF padding byte decodes as a DLC of −1. The offset then moved backwards over a padded tail, rescanning the frame at misaligned positions and inventing contained PDUs out of padding. Both header fields are now read unsigned. This one is invisible when `canmatrix.Frame.unpack` is the only oracle, because it makes the same assumption.

Results after the fixes:

| check | result |
|---|---|
| vs `canmatrix.Frame.unpack`, sample by sample | 1 237 247 values, **0 mismatches** |
| vs the signal-based recording of the same drive | 298 of 303 signals agree on **every** one of 114 251 samples |
| non-container signals, `master` vs this branch | 541 signals **bit-identical**, 0 changed |
| container signals gained | 789 (master yields only `Header_ID`/`Header_DLC`) |

The five signals in row 2 that do not agree everywhere are free-running sequence counters and CRCs whose sample instants differ between the two recordings by more than the comparison window — identical value range and cardinality.

Two regression tests were added; defect 3 is covered by the existing oracle test now that it asserts absolute values. Full description updated above.

One prerequisite fell out of this: reading these measurements at all needs #1307. `CAN_DataFrame.DataBytes` is a VLSD channel inside the composed `CAN_DataFrame` structure, and on `master` its signal data cannot be located, so `extract_bus_logging` raises before it ever reaches this code. That fix is independent of this PR and is submitted separately.

kipp-ing:
PR adjusted for easier review. pls let me know if I can help further.

danielhrisca (approved):

danielhrisca:
thanks

Issue #1313, opened 2026-09-10
WITH_SYSTEM_DEFLATE=ON cannot configure or link: wrong CMake package name for zstd, and vendored target names used for the system libraries


sporksnail:
**Affected versions:** 8.8.0 through 8.8.27 (current). 8.7.2 and earlier are fine - they only needed system libdeflate.

## Summary

Building with `-DWITH_SYSTEM_DEFLATE=ON` fails. The `WITH_SYSTEM_DEFLATE` branch of `CMakeLists.txt` asks for a CMake package called `libzstd`, which doesn't exist, and then links against `lz4_shared` / `libzstd_shared` - target names that only exist in the `else()` branch, where `ext/lz4` and `ext/zstd` are pulled in with `add_subdirectory()`.

This matters for distribution packaging: `ext/libdeflate`, `ext/lz4` and `ext/zstd` are git submodules, so they're empty in the GitHub release tarballs, and `WITH_SYSTEM_DEFLATE=ON` is the only way to build the `cutils` extension from a released sdist/tarball. It looks like this path isn't covered by CI - I couldn't find `WITH_SYSTEM_DEFLATE` set anywhere under `.github/` or in `pyproject.toml`, which would explain how it went unnoticed.

## Reproduction

On a system with development packages for libdeflate, lz4 and zstd installed (here Debian unstable: `libdeflate-dev`, `liblz4-dev`, `libzstd-dev`, CMake 4.3.4):

```
cmake -S . -B build -DWITH_SYSTEM_DEFLATE=ON
```

### Problem 1 - `find_package(libzstd)` finds nothing

```
CMake Error at CMakeLists.txt:27 (find_package):
  By not providing "Findlibzstd.cmake" in CMAKE_MODULE_PATH this project has
  asked CMake to find a package configuration file provided by "libzstd", but
  CMake did not find one.
```

zstd installs its config file as `zstdConfig.cmake`, so the package is named **`zstd`**, not `libzstd`. From zstd's own `build/cmake/CMakeModules/ZstdPackage.cmake`:

```cmake
configure_package_config_file(zstdConfig.cmake.in "${CMAKE_CURRENT_BINARY_DIR}/zstdConfig.cmake" ...)
install(EXPORT "zstdExports${target_suffix}" ... NAMESPACE zstd::)
```

(`find_package(lz4)` is already correct - lz4 does install `lz4Config.cmake`.)

### Problem 2 - the linked target names don't exist either

[...]

PR #1314, opened 2026-09-10
Use the CMake package and target names the system lz4/zstd export

sporksnail:
Fixes: #1313 

 The WITH_SYSTEM_DEFLATE branch looks for a CMake package called "libzstd"
 and links against the bare targets "lz4_shared" and "libzstd_shared".
 Those names only exist in the else() branch, where the vendored
 ext/lz4 and ext/zstd subprojects are pulled in with add_subdirectory().
 
 The installed lz4 and zstd both export a config package and namespaced
 imported targets instead:
 
```
lz4Config.cmake   -> LZ4::lz4_shared, LZ4::lz4_static, LZ4::lz4
zstdConfig.cmake  -> zstd::libzstd_shared, zstd::libzstd_static,
                     zstd::libzstd
 ```

 Use those, matching the libdeflate::libdeflate_shared already on the same
 line.


danielhrisca (approved):

danielhrisca:
thanks

PR #1315, opened 2026-09-18
fix: repair the tests workflow (FileWidget str paths, Windows build, stale test API)

kipp-ing:
The `tests` workflow fails on `development` for several independent reasons. This fixes the ones with a clear cause. The mypy job still fails on Python 3.10 and 3.13, so the `tests` check as a whole stays red there (see below).

- **`FileWidget` with a `str` path.** `FileWidget.__init__` read `file_name.suffix` before converting `file_name` to a `Path`, so every GUI test that passes a `str` (the test helper does, via `shutil.copy`) failed with `AttributeError: 'str' object has no attribute 'suffix'`. The regression came in with #1301. The `Path` conversion now happens first.
- **Stale test API.** `test_FileWidget_TabModifyAndExport` used the `QListWidget` API (`.count()`, `.item()`) on `selected_filter_channels`, which is a `QTreeWidget` in the file widget. It was hidden behind the `suffix` error. It now uses `topLevelItemCount()` / `topLevelItem()` / `text(0)`.
- **Windows test jobs.** They failed building the sdist with `CMAKE_C_COMPILER not set` (CMake picks the NMake generator). Same setup steps as `pyinstaller_build` from #1310 (Ninja + Visual Studio 2022 Build Tools). With this the Windows jobs get through the build and run the test suite. Some of these steps may turn out to be redundant and can be trimmed afterwards.
- **`ruff format --check`** failed on `gui/utils.py` (trailing whitespace only).

## Result on this PR

- pytest passes on every job that gets to run it: 223 passed on Linux, 225 passed on Windows 3.10 / 3.12 / 3.13. `ruff` and `doc` are OK.
- Windows 3.11 was killed by the 600 s pytest timeout in `test_PlotWidget_ContextMenu::test_Action_CopyChannelStructure_Group`, blocked in `drag.exec()` (`gui/widgets/tree.py`). The same test passes on the other three Windows jobs in the same run, so this looks like a flaky drag-and-drop hang unrelated to this change.

## Not addressed

- The mypy job: ~430 errors on 3.10, exit code 2 on 3.13 (checked on Linux and Windows). It is the only thing still failing on those jobs.


danielhrisca (approved):

danielhrisca:
thanks @kipp-ing

PR #1316, opened 2026-09-23
fix reading of array signals

Simsalabim1234:
The reading of channels that contain array data is fixed.


Simsalabim1234:
Hi @danielhrisca , kindly request to merge this PR to development branch. It fixes that reading of signals that have array structures. Currently this does not work on development and master branch.

